PR #7, opened 2014-11-13
Add the libselinux-python package for RedHat.

benjifisher:
When I switched the base box to chef/centos-6.6, I got the error "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!" and this fixes it.


geerlingguy:
@benjifisher - That particular package should be installed prior to the role being run (maybe in `pre_tasks` or in a separate setup role), since it isn't directly related to MySQL.

Issue #11, opened 2014-12-12
Problem restarting mysql service after configuration applied


CharlyP:
Hi,

I'm currently using your role on a Debian wheezy (with vagrant and LXC).
I only set `mysql_databases` and `mysql_users` vars but at the end of the provisioning I get the following error:
```
failed: [default] => {"failed": true}
msg: Starting MySQL database server: mysqld . . . . . . . . . . . . . . failed!
```

The corresponding syslog logs are:
```
Dec 12 06:13:46 corep /etc/init.d/mysql[11915]: 0 processes alive and '/usr/bin/mysqladmin --defaults-file=/etc/mysql/debian.cnf ping' resulted in
Dec 12 06:13:46 corep /etc/init.d/mysql[11915]: #007/usr/bin/mysqladmin: connect to server at 'localhost' failed
Dec 12 06:13:46 corep /etc/init.d/mysql[11915]: error: 'Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)'
Dec 12 06:13:46 corep /etc/init.d/mysql[11915]: Check that mysqld is running and that the socket: '/var/run/mysqld/mysqld.sock' exists!
Dec 12 06:13:46 corep /etc/init.d/mysql[11915]:
```

Any idea what went wrong during configuration?

mb720:
In the meantime I could get around the issue by setting `mysql_innodb_log_file_size: "5M"`

geerlingguy:
Recent commits that are now in the 1.4.1 release should resolve the issues. The role was working on some installations of Ubuntu (both 12.04 and 14.04), but only due to some particular bits of environment configuration. I've updated the role to work much better cross-platform, and I've also added tests to ensure socket-based connection and TCP-based connection works correctly.

mb720:
Thanks for the fix.
I had to delete ib_logfile0 and ib_logfile1 in /var/lib/mysql/ to make provisioning work using 1.4.1.

moreirapontocom:
After spending about 5 hours working on that, you saved my life, mb720. Thanks a lot.

PR #12, opened 2014-12-29
Adhere to the Debian idioms more closely for some config values.

jimbocoder:
I'm not even sure this role works properly on Debian/Ubuntu currently, because the init scripts look for `/etc/mysql/my.cnf` but the role actually writes to `/etc/my.cnf`.  This should address that, among with some other minor things.

`/etc/mysql/my.cnf` by default does an "include" of `/etc/mysql/conf.d/*.cnf` so rather than writing the main config file, I changed the Debian version of the role to write to the `conf.d/` directory.

Finally, I changed the default `mysql_socket` value based on the distro.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -2,7 +2,7 @@
 - name: Copy my.cnf global MySQL configuration.
   template:
     src: my.cnf.j2
-    dest: /etc/my.cnf
+    dest: "{{ mysql_config_file_path }}"
     owner: root
     group: root
     mode: 0644
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -3,3 +3,5 @@ mysql_daemon: mysql
 mysql_packages:
   - mysql-server
   - python-mysqldb
+mysql_config_file_path: /etc/mysql/conf.d/ansible-overrides.cnf
+mysql_socket: /var/run/mysqld/mysqld.sock
--- a/vars/RedHat.yml
+++ b/vars/RedHat.yml
@@ -4,3 +4,5 @@ mysql_packages:
   - mysql
   - mysql-server
   - MySQL-python
+mysql_config_file_path: /etc/my.cnf
+mysql_socket: /var/lib/mysql/mysql.sock


mrconnerton:
I thought I had the same problem so I applied the patch but I'm still having issues starting mysql. Here is a snippit from my log:

```
150106 18:49:04 [Note] /usr/sbin/mysqld: Shutdown complete

150106 18:49:04 mysqld_safe mysqld from pid file /var/run/mysqld/mysqld.pid ended
150106 18:50:04 mysqld_safe Starting mysqld daemon with databases from /var/lib/mysql
150106 18:50:04 [Warning] Using unique option prefix key_buffer instead of key_buffer_size is deprecated and will be removed in a future release. Please use the full name instead.
150106 18:50:04 [Warning] 'THREAD_CONCURRENCY' is deprecated and will be removed in a future release.
150106 18:50:04 [Warning] Using unique option prefix myisam-recover instead of myisam-recover-options is deprecated and will be removed in a future release. Please use the full name instead.
150106 18:50:04 [Note] Plugin 'FEDERATED' is disabled.
150106 18:50:04 InnoDB: The InnoDB memory heap is disabled
150106 18:50:04 InnoDB: Mutexes and rw_locks use GCC atomic builtins
150106 18:50:04 InnoDB: Compressed tables use zlib 1.2.7
150106 18:50:04 InnoDB: Using Linux native AIO
150106 18:50:04 InnoDB: Initializing buffer pool, size = 256.0M
150106 18:50:04 InnoDB: Completed initialization of buffer pool
InnoDB: Error: log file ./ib_logfile0 is of different size 0 5242880 bytes
InnoDB: than specified in the .cnf file 0 67108864 bytes!
150106 18:50:04 [ERROR] Plugin 'InnoDB' init function returned error.
150106 18:50:04 [ERROR] Plugin 'InnoDB' registration as a STORAGE ENGINE failed.
150106 18:50:04 [ERROR] Unknown/unsupported storage engine: InnoDB
150106 18:50:04 [ERROR] Aborting
```

jimbocoder:
@mrconnerton If you don't intend to use replication at all, you can disable the replication-related settings this role includes (`server-id`, etc) and set the log file size to 5242880 as that appears to be what it was previous to running the role.

mrconnerton:
Thanks @jimbocoder  I will give that a try.

geerlingguy:
Sorry I didn't see this earlier—I've been working on making the role work with Ubuntu 14.04 and 12.02 this week, and only *after* started committing fixes did I notice the PR. D'oh!

I'm going to move conversation back into #11 and close this PR; I'm hoping the changes I've made will make the role work as well on Debian/Ubuntu as on RedHat/CentOS. Versions 1.4.1+ should hopefully solve the issues here and in #11.

mrconnerton:
Thanks @geerlingguy, settings of mysql_server_id: "" and mysql_innodb_log_file_size: "5242880" and your latest commit I assume makes it provision perfectly.

PR #13, opened 2014-12-29

jimbocoder:
I use a special sudo user to run ansible, rather than a root user.  Since Ansible modules like _mysql_user_ read credentials from `$HOME/.my.cnf` but the role tasks could be in sudo mode, the modules are always trying to read from `/root/.my.cnf` regardless of the actual ansible user or `mysql_user_home` rolevar.  This works great when you happen to be using the root user anyway, but otherwise, problems.

This hopes to address that.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
-mysql_user_home: /root
+mysql_user_home: "~{{ ansible_ssh_user }}"
 mysql_root_username: root
 mysql_root_password: root
 
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -5,4 +5,5 @@
     collation: "{{ item.collation | default('utf8_general_ci') }}"
     encoding: "{{ item.encoding | default('utf8') }}"
     state: present
+  sudo: no
   with_items: mysql_databases
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ None.
 
 Available variables are listed below, along with default values (see `vars/main.yml`):
 
-    mysql_user_home: /root
+    mysql_user_home: <home directory of the ansible user>
 
-The home directory inside which Python MySQL settings will be stored, which Ansible will use when connecting to MySQL. This should be the home directory of the user which runs this Ansible role.
+The home directory inside which Python MySQL settings will be stored, which Ansible will use when connecting to MySQL. This should be the home directory of the user which runs this Ansible role, which the default value should take care of automatically.
 
     mysql_root_password: root
 
--- a/tasks/replication.yml
+++ b/tasks/replication.yml
@@ -6,6 +6,7 @@
     password: "{{ mysql_replication_user.password }}"
     priv: "{{ mysql_replication_user.priv | default('*.*:REPLICATION SLAVE') }}"
     state: present
+  sudo: no
   when: >
     (mysql_replication_role == 'master')
     and mysql_replication_user
@@ -13,6 +14,7 @@
 
 - name: Check slave replication status.
   mysql_replication: mode=getslave
+  sudo: no
   ignore_errors: true
   register: slave
   when: >
@@ -21,6 +23,7 @@
 
 - name: Check master replication status.
   mysql_replication: mode=getmaster
+  sudo: no
   delegate_to: "{{ mysql_replication_master }}"
   register: master
   when: >
@@ -36,6 +39,7 @@
     master_password: "{{ mysql_replication_user.password }}"
     master_log_file: "{{ master.File }}"
     master_log_pos: "{{ master.Position }}"
+  sudo: no
   ignore_errors: True
   when: >
     slave|failed
@@ -45,6 +49,7 @@
 
 - name: Start replication.
   mysql_replication: mode=startslave
+  sudo: no
   when: >
     slave|failed
     and (mysql_replication_role == 'slave')
--- a/tasks/secure-installation.yml
+++ b/tasks/secure-installation.yml
@@ -10,6 +10,7 @@
     name: "root"
     host: "{{ item }}"
     password: "{{ mysql_root_password }}"
+  sudo: no
   with_items:
     - 127.0.0.1
     - ::1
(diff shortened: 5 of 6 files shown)


defaults/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-mysql_user_home: /root
+mysql_user_home: "~{{ ansible_ssh_user }}"
```
geerlingguy: Shouldn't this just be `"~"`?

tasks/databases.yml
```diff
@@ -5,4 +5,5 @@
     collation: "{{ item.collation | default('utf8_general_ci') }}"
     encoding: "{{ item.encoding | default('utf8') }}"
     state: present
+  sudo: no
```
geerlingguy: I'd rather not set `sudo: no` for all these tasks, but be more explicit about which (if any) tasks truly _must_ be run as sudo (e.g. the `apt` or `yum` tasks). That way, for the minority of playbooks which are fully run without `sudo: yes`, the adding `sudo: no` is redundant.

defaults/main.yml
jimbocoder: That's actually what I initially tried.  And it works in almost every case, since it's the same thing.  The problem I ran into was, e.g. running the playbook as a `vagrant` user, but the `ansible` user initially bootstrapped the mysql installation, so only `ansible` has the credentials file.  This hopefully accounts for the situation where `ansible_ssh_user` has been explicitly set in `ansible.cfg` (which was my situation).

tasks/databases.yml
jimbocoder: Agreed, it's kind of unwieldy, but still seems prudent to me.  In a given playbook probably most tasks don't really need sudo, but it seems like people tend to run them with `sudo: yes` just to make life easy.  I believe it's helpful to be explicit here so the role runs the same whether it's invoked with or without sudo.  I could go either way on that, let me know if you'd still like it changed.

geerlingguy:
See also: #36 .

defaults/main.yml
solomongifford: We've seen this behavior as well.

tasks/databases.yml
solomongifford: We also use another user, but we don't log in as that user with ansible, so "sudo: no" would fail in our case.

drzraf:
See also: #100

solomongifford:
We solved this a different way - 

We assume that because mysql has a default "root" user named root, that the real question is not "what is the root user's name?" but "is there a second user beside root?"

This creates a few variables like the following:
```
---
mysql_root_home: /root
mysql_root_username: root
mysql_root_password: root
mysql_default_user_home: /root
mysql_default_username: root
mysql_default_password: root
```
This implies we change our secure installation to be:
```
---
- name: Ensure default user is present.
  mysql_user:
    name: "{{ mysql_default_username }}"
    host: 'localhost'
    password: "{{ mysql_default_password }}"
    priv: '*.*:ALL,GRANT'
    state: present
  when: mysql_default_username != mysql_root_username

# Has to be after the password assignment, for idempotency.
- name: Copy user-my.cnf file with password credentials.
  template:
    src: "user-my.cnf.j2"
    dest: "{{ mysql_default_user_home }}/.my.cnf"
    owner: "{{ mysql_default_username }}"
    mode: 0600
  when: mysql_default_username != mysql_root_username and (mysql_install_packages | bool or mysql_default_password_update)

- name: Disallow root login remotely
  command: 'mysql -NBe "{{ item }}"'
  with_items:
    - DELETE FROM mysql.user WHERE User='{{ mysql_root_username }}' AND Host NOT IN ('localhost', '127.0.0.1', '::1')
  changed_when: False

- name: Get list of hosts for the root user.
  command: mysql -NBe "SELECT Host FROM mysql.user WHERE User = '{{ mysql_root_username }}' ORDER BY (Host='localhost') ASC"
  register: mysql_root_hosts
  changed_when: false
  when: mysql_install_packages | bool or mysql_root_password_update

# Note: We do not use mysql_user for this operation, as it doesnt always update
# the root password correctly. See: https://goo.gl/MSOejW
- name: Update MySQL root password for localhost root account.
  shell: mysql -NBe 'SET PASSWORD FOR "{{ mysql_root_username }}"@"{{ item }}" = PASSWORD("{{ mysql_root_password }}")'
  with_items: "{{ mysql_root_hosts.stdout_lines }}"
  when: mysql_install_packages | bool or mysql_root_password_update

# Has to be after the root password assignment, for idempotency.
- name: Copy .my.cnf file with root password credentials.
  template:
    src: "root-my.cnf.j2"
    dest: "{{ mysql_root_home }}/.my.cnf"
    owner: root
    group: root
    mode: 0600
  when: mysql_install_packages | bool or mysql_root_password_update

- name: Get list of hosts for the anonymous user.
  command: mysql -NBe 'SELECT Host FROM mysql.user WHERE User = ""'
  register: mysql_anonymous_hosts
  changed_when: false

- name: Remove anonymous MySQL users.
  mysql_user:
     name: ""
     host: "{{ item }}"
     state: absent
  with_items: "{{ mysql_anonymous_hosts.stdout_lines }}"

- name: Remove MySQL test database.
  mysql_db: "name='test' state=absent"
```

solomongifford:
I'll add, this theoretically could mean that  mysql_root_home and mysql_root_username are now vars, not defaults. 

If you like this approach let me know and I can do a PR.

drzraf:
Do we all keep in mind that the day this role switch to Ansible 2.0+ will we (very happily) throw away all these workarounds attempt and use the `mysql_(user|db)` `config_file` parameter, isn't?

jimbocoder:
As the OP, none of the responses here fully solve the initial perceived deficiency, and at the same time my suggestions do not address all of the responses. Since this issue is getting a bit old, and because the upgrade path between Ansible 1.x and 2.0 is really quite smooth (compared to almost any other increment in the major version number) my personal suggestion is to concur with @drzraf and push for 2.0 compliance.

geerlingguy:
I'm inclined to go fully with 2.0 as well; it's been long enough and all my infra playbooks work fine on 2.x now that a few pernicious bugs in the first/second point releases have been fixed. It will make this role simpler and offload some of the weird logic to Ansible's own modules.

geerlingguy:
Closing in favor of https://github.com/geerlingguy/ansible-role-mysql/pull/13

PR #21, opened 2015-04-05
Tweaks for mariadb usage

mwanjura:


mwanjura:
Oh, sorry. Just saw that travis failes. I'll fix it and request a new pull request.

defaults/main.yml
```diff
@@ -20,6 +20,8 @@ mysql_enablerepo: ""
 mysql_port: "3306"
 mysql_bind_address: '0.0.0.0'
 mysql_datadir: /var/lib/mysql
+mysql_daemon: mysqld
```
geerlingguy: This won't work correctly (by default) for Debian/Ubuntu users. For convenience's sake, we could use the `__varname` shortcut as seen in my PHP role: https://github.com/geerlingguy/ansible-role-php/blob/master/vars/Debian.yml#L15 (that way we set a default dynamically if the user doesn't supply an override).

geerlingguy:
Sorry about forgetting about this PR—I chose to use the approach in #38, which works across CentOS/RHEL and Ubuntu/Debian.

Issue #23, opened 2015-04-09
Root password for hostname for local_action


solomongifford:
Related to #22, if local_action is used with a mysql install ansible_hostname is localhost and not the hostname of the machine and the root password is not set.

Suggestion would be to grab the results of:

command: 'mysql -NBe 'SELECT Host from mysql.user WHERE User = "root" order by (Host="localhost") ASC'

Note that this orders the localhost last so that ansible doesn't lock itself out.

geerlingguy:
Nice. Silly me, trying to manipulate the result set with Python/Jinja2... when just adding the order statement to the MySQL statement would work perfectly.

This should be simple enough to do—thanks for reporting, and I'll try to get to this soon!

geerlingguy:
Fixed in PR #28.

PR #25, opened 2015-04-09
Support for removing incompatible packages, optional my.cnf, issues #23, #26, etc.

solomongifford:
Sometimes during an install there are packages that are already installed that first must be removed before mysql can be installed.  The first two commits represent that scenario.

The third commit supports a "set and forget" strategy for the global my.cnf, allowing scenarios where you want future runs to not update the my.cnf

The fourth commit fixes issue #23.

The fifth commit fixes issue #26.

The sixth/seventh commit adds support for https://github.com/ansible/ansible-modules-core/pull/830

The eighth commit is a simple rename of the python-my.cnf.j2 to user-my.cnf.j2 to more accurately reflect its purpose.

The ninth & tenth commits put the performance variables into an array so that other variables can also be used by adding rows to the array (or removing in order to apply defaults).


solomongifford:
These changes have been tested on CentOS 6 both for first runs and repeat runs. Thanks for your consideration.

README.md
```diff
@@ -43,6 +39,19 @@ The MySQL users and their privileges. A user has the values `name`, `host` (defa
 
 (RedHat/CentOS only) If you have enabled any additional repositories (might I suggest geerlingguy.repo-epel or geerlingguy.repo-remi), those repositories can be listed under this variable (e.g. `remi,epel`). This can be handy, as an example, if you want to install later versions of MySQL.
 
+    mysql_incompatible_packages:
```
geerlingguy: I'd rather not do this as part of the `geerlingguy.mysql` role; in general, if there are packages that shouldn't be installed on a given server, or other setup tasks outside the purview of MySQL configuration, a `pre_task` or another setup role that runs prior to this role should do it.

As an example, I make sure to always use minimal base boxes for CentOS/RHEL/Debian/Ubuntu, which allow me to build up on top of them only what's needed. If the base image you use with Vagrant, DigitalOcean, AWS, etc. has extra cruft, you need to switch to a different, more minimalist image, or add a task to remove the packages prior to this role.

geerlingguy:
Most of these changes are simple enough and I'd like to merge them staight away, but some of them are pretty far-reaching and would at least warrant a major version bump to 2.0.

If you could break each feature/change into a separate PR, it would be easier to pull in the simpler ones and hash out the more far-reaching changes :)

solomongifford:
Done, with the exception of incompatible packages.  Also, in the last commit I had also changed the order of some of the items in the readme to match the order in defaults/main.yml but doing that in a separate PR would cause conflicts with other of my PR's.

PR #27, opened 2015-04-10
optional-mycnf-force

solomongifford:
Supports a "set and forget" strategy for the global my.cnf, allowing scenarios where you want future runs to not update the my.cnf


geerlingguy:
I don't think this does what is intended in the description above... setting `force: no` would only apply to symlink creation, according to the [Ansible docs for `file`](http://docs.ansible.com/file_module.html#options).

We would need a different variable, like `mysql_use_managed_mycnf` or something like that, similar to how my `geerlingguy.php` role has a `php_use_managed_ini` variable ([example](https://github.com/geerlingguy/ansible-role-php/blob/master/defaults/main.yml#L21). If it's set to true, current behavior remains. If set to false, it's assumed you will manage the my.cnf file through your own means.

geerlingguy:
Well, actually... looks like this is a hole in the docs—I guess `template` inherits the `force` behavior from `copy`, not `file` (don't know why I was assuming that). The way you have it configured could work, but I wonder—if you aren't using this role to continue managing the MySQL installation, is this role just being used to build a base image or something, then other playbooks are doing further my.cnf configuration?

Or—gasp—are you managing my.cnf files by hand? ;-)

solomongifford:
*gasp* is right - but only kinda. In our scenario we use ansible for installation and maintenance runs but allow clients to modify their my.cnf's for custom tuning.  We essentially set the default and let them stray from there.

solomongifford:
I'll add - We've been using the force=no on the my.cnf on our ansible runs for a few months now.

geerlingguy:
@solomongifford - I understand on the _gasp_—I forgot that your use case is for user-managed hosting.

I'll sit on this a couple days and try to decide which way I'd like to go (either allowing management or not, like in the PHP role, or allowing `force: no`)... I'm leaning towards your implementation, but my brain is getting fried this afternoon, and all your simple/easy win PRs are already merged :)

PR #31, opened 2015-04-10
Allow performance configurations of variables not currently supported

solomongifford:
Per your comment on https://github.com/geerlingguy/ansible-role-mysql/pull/25, I made this "1.0 compatible."  Ultimately I think it makes sense to put all the performance defaults into the arrays.  Let me know if you want me to update the pr.


geerlingguy:
I think this should work. Going to do a little more testing and let it sit so I can see if the addition sits well.

solomongifford:
As an aside, you have max_allowed_packet in your settings file twice - if nothing else, this demonstrates that if the setting is in the config twice it shouldn't break things.

geerlingguy:
D'oh! Right you are :)

PR #32, opened 2015-04-10
added support for update_password=on_create for ansible_version 2.0+

solomongifford:


geerlingguy:
@solomongifford - Is this going to be available in 1.9.x? I notice it was merged into the Ansible `devel` branch, and has `version_added: "1.9"` in the PR...

solomongifford:
There's a commit right after that changing the release to 2.0.  The merge request was made before 1.9 but accepted after 1.9.

defaults/main.yml
```diff
@@ -66,7 +66,8 @@ mysql_databases: []
 mysql_users: []
 # Full Example:
 # mysql_users:
-#   - { name: example, host: 127.0.0.1, password: secret, priv: *.*:USAGE }
+#   - { name: example, host: 127.0.0.1, password: secret, update_password: allways,  priv: "*.*:USAGE" }
```
JorisVanEijden: 'allways' should be 'always'

tasks/users.yml
```diff
@@ -7,3 +7,15 @@
...
+  when: ansible_version.full | version_compare('2.0', '<')
+
+- name: Ensure MySQL users are present.
+  mysql_user:
+    name: "{{ item.name }}"
+    host: "{{ item.host | default('localhost') }}"
+    password: "{{ item.password }}"
+    priv: "{{ item.priv | default('*.*:USAGE') }}"
+    state: present
+    update_password: "{{ item.update_password }}"
```
JorisVanEijden: No default provided.

solomongifford:
This is likely no longer an issue due to the fact the module no longer executes the mysql_user after the first run.

PR #43, opened 2015-07-06
Disallow remote mysql root connections

kostyrevaa:
This PR adds feature to disallow remote mysql root connections so we can fight mysql root brutforcing.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -3,6 +3,8 @@ mysql_user_home: /root
 mysql_root_username: root
 mysql_root_password: root
 
+mysql_disallow_remote_root: false
+
 mysql_enabled_on_startup: yes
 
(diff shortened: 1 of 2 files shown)


kostyrevaa:
Up

geerlingguy:
@kostyrevaa - Sorry about not taking a look at this sooner—this is definitely something I want to add; I've merged the similar change from #75, which is a little more secure in the way it handles the deletion (since the fqdn might be different than what's in the database already).

kostyrevaa:
yeah.
It took you six months.
Fortunately for me I use mysql no longer.

geerlingguy:
@kostyrevaa - Sorry about the delay, but I'm glad you're using something that suits your needs now!

PR #44, opened 2015-07-13
Replace Ubuntu init.d script

jwhitlock:
This fixes random issues running provisioning for a Vagrant VM with Ubuntu 14.04 and mysql-community-server 5.6.25-1ubuntu12.04.

/etc/init.d/mysql greps for the mysql process to determine if mysql is running. When running, the process count includes the mysqld process and the grep for that process (count=2). When running under vagrant using ansible to provision, 5% - 10% of the time the grep is not in the process list, causing false negatives when detecting the mysqld process is running. When provisioning, this results in refusing to stop the service and failing to start the service.

This new init script ignores the grep process, making the run detection more reliable.

Submitted upstream as http://bugs.mysql.com/bug.php?id=77696

### DIFF
--- a/files/etc/init.d/mysql
+++ b/files/etc/init.d/mysql
@@ -1,0 +1,182 @@
+#!/bin/bash
+#
+### BEGIN INIT INFO
+# Provides:          mysql
+# Required-Start:    $remote_fs $syslog
+# Required-Stop:     $remote_fs $syslog
+# Should-Start:      $network $time
+# Should-Stop:       $network $time
+# Default-Start:     2 3 4 5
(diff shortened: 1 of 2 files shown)


geerlingguy:
I'd rather not be in the business of maintaining parallel/bugfix init scripts in this Ansible role if I can avoid it—the role is reliable in my usage with the default versions of MySQL included with either 12.04 or 14.04, and also works fine in my testing with MariaDB on 14.04, so I'd rather leave things be for now, and hope the upstream bug is fixed so this is not necessary in your forked version anymore, either :)

PR #46, opened 2015-08-07
Add ability to install the MySQL client only

ChineduUzoka:
You wrote a very good role that im using.  I needed the ability to install the client only but managed via a role then install independently


ChineduUzoka:
Tests need mysql-server installed whereas my changes has a mysql client only install

PR #50, opened 2015-08-20
eliminates extra whitespace in the variable value

jpeak5:
On ubuntu 14.04, removing this extra whitespace cleared an error that was halting ansible at line 25 of tasks/configure.yml.


geerlingguy:
Thanks!

PR #65, opened 2015-10-21
Allow append privileges

raistlin:
I need one user to access two different databases, with this change I can handle it.

### DIFF
--- a/tasks/users.yml
+++ b/tasks/users.yml
@@ -6,4 +6,5 @@
     password: "{{ item.password }}"
     priv: "{{ item.priv | default('*.*:USAGE') }}"
     state: present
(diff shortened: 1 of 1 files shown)


solomongifford:
Have you tested doing something like:

```
mysql_users:
  - name: example_user
    host: "%"
    password: similarly-secure-password
    priv: "example_db.*:ALL"
  - name: example_user
    host: "%"
    password: similarly-secure-password
    priv: "example2_db.*:ALL"
```

geerlingguy:
This could still be helpful as a general addition, right? I haven't personally needed to use append_privs, but it could be useful if someone needs it.

PR #70, opened 2015-11-12
make connection count adjustable

caseycs:


geerlingguy:
Simple and works great for me!

PR #72, opened 2015-11-24
Add variables for innodb large prefix

skuroki:
For utf8mb4 encoding. see https://www.pagerduty.com/blog/trading-up-your-engine-how-to-move-your-iops-heavy-mysqlrails-stack-to-unicode-without-downtime/


solomongifford:
I think this is better solved with  PR #31

geerlingguy:
Closing in favor of #76 - Thanks for submitting the PR @skuroki ! #76 allows this to be resolved by simply providing the variables in your own .cfg file to included with MySQL's configs.

PR #74, opened 2015-12-11
reversed order of mysql-python install

solomongifford:
MySQL-python has a dependency of mysql-libs.  If MySQL-python is installed before mysql, mysql-libs will be installed. However, when installing mysql56u (for example) and other versions of mysql, a mysql56u-libs is installed which conflicts with mysql-libs.  

By installing mysql(version) first, mysql(version)-libs will satisfy the requirement for MySQL-python and no conflict arises.

This PR just swaps the order of the install so that MySQL-python is installed after the packages.


geerlingguy:
Simple and makes sense to me!

PR #76, opened 2015-12-15
Allow configuration of variables not currently supported

solomongifford:
This is the same as PR #31 but that repo was moved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Whether the global my.cnf should be overwritten each time this role is run. Sett
 
     mysql_config_include_files: []
 
-A list of files that should override the default global my.cnf.   Each item in the array requires a "src" parameter which is a path to a file.  An optional "force" parameter can force the file to be updated each time ansible runs.
+A list of files that should override the default global my.cnf. Each item in the array requires a "src" parameter which is a path to a file. An optional "force" parameter can force the file to be updated each time ansible runs.
 
     mysql_databases: []
 
--- a/README.md
+++ b/README.md
@@ -75,6 +75,13 @@ Slow query log settings. Note that the log file will be created by this role, bu
 
 The rest of the settings in `defaults/main.yml` control MySQL's memory usage. The default values are tuned for a server where MySQL can consume ~512 MB RAM, so you should consider adjusting them to suit your particular server better.
 
+    mysqld_performance_settings: 
+     - { name: key_buffer_size, value: "256M" }
+    mysql_dump_settings: []
+     - { name: mysqldump_max_allowed_packet, value: "64M" }
+
+Any additional performance settings you would like to add beyond the defaults.
+
     mysql_server_id: "1"
     mysql_max_binlog_size: "100M"
     mysql_expire_logs_days: "10"
--- a/README.md
+++ b/README.md
@@ -32,6 +32,10 @@ Whether MySQL should be enabled on startup.
 
 Whether the global my.cnf should be overwritten each time this role is run. Setting this to `no` tells Ansible to only create the `my.cnf` file if it doesn't exist. This should be left at its default value (`yes`) if you'd like to use this role's variables to configure MySQL.
 
+    mysql_config_include_files: []
+
+A list of files that should override the default global my.cnf.   Each item in the array requires a "src" parameter which is a path to a file.  An optional "force" parameter can force the file to be updated each time ansible runs.
+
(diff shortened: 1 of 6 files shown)


solomongifford:
I'm actually going to modify this PR - don't pull yet.

geerlingguy:
Sounds good.

solomongifford:
OK, this updated PR goes in a completely different direction and solves another problem as well.

Mysql implementations support using the includedir directive which allows files to override settings in in the my.cnf.  Mariadb does this by default.  

This PR allows the files in the default /etc/mysql/conf.d (Debian based) or /etc/my.cnf.d (CentOS based) to be managed by this ansible role.  Instead of managing the contents of the files, it allows the override files to be managed by the role.   

This PR both solves the need to support more options (as they can be simply added to an override file) and the need to manage the mariadb cnf files.

README.md
```diff
@@ -32,6 +32,10 @@ Whether MySQL should be enabled on startup.
 
 Whether the global my.cnf should be overwritten each time this role is run. Setting this to `no` tells Ansible to only create the `my.cnf` file if it doesn't exist. This should be left at its default value (`yes`) if you'd like to use this role's variables to configure MySQL.
 
+    mysql_config_include_files: []
+
+A list of files that should override the default global my.cnf.   Each item in the array requires a "src" parameter which is a path to a file.  An optional "force" parameter can force the file to be updated each time ansible runs.
```
geerlingguy: Can you change to one space between period/new sentence?

defaults/main.yml
```diff
@@ -70,6 +70,13 @@ mysql_log: ""
 mysql_log_error: /var/log/mysql.err
 mysql_syslog_tag: mysql
 
+mysql_config_include_files: []
+# Full example:
```
geerlingguy: Just to follow the example set in many of my other roles, you can delete this line and the following, and just have the following two lines (with the actual example) commented, below the empty (`[]`) var.

defaults/main.yml
```diff
@@ -70,6 +70,13 @@ mysql_log: ""
...
+# mysql_config_include_files:
+#   - src: path/relative/to/playbook/file.cnf
+#   - { src: path/relative/to/playbook/anotherfile.cnf, force: yes }
+
```
geerlingguy: Just one space.

tasks/configure.yml
```diff
@@ -9,6 +9,26 @@
...
+  when: mysql_config_include_files|length
+
+- name: Copy my.cnf override files into include directory.
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ mysql_config_include_dir }}/{{ item.src | basename }}"
+    owner: root
+    group: root
+    mode: 0644
+    force: "{{ item.force|default(False) }}"
```
geerlingguy: please include a space between the brackets.

templates/my.cnf.j2
```diff
@@ -96,3 +96,10 @@ max_allowed_packet = {{ mysql_mysqldump_max_allowed_packet }}
 
 [mysqld_safe]
 pid-file = {{ mysql_pid_file }}
+
+{% if mysql_config_include_files|length %}
+# * IMPORTANT: Additional settings that can override those from this file!
+#   The files must end with '.cnf', otherwise they'll be ignored.
+#
+!includedir {{ mysql_config_include_dir }}
+{% endif %}
```
geerlingguy: needs newline

templates/my.cnf.j2
```diff
@@ -96,3 +96,10 @@ max_allowed_packet = {{ mysql_mysqldump_max_allowed_packet }}
 
 [mysqld_safe]
 pid-file = {{ mysql_pid_file }}
+
+{% if mysql_config_include_files|length %}
```
geerlingguy: Please add space around |

geerlingguy:
I really like the approach; it offers as much flexibility as the role consumer needs, and is much nicer than blindly adding hundreds more options to the config.

defaults/main.yml
solomongifford: Sure - I assume you'll want the other examples in this cleaned up as well, but I'll leave that for another day.

solomongifford:
Let me know if I caught all the spacing changes.

geerlingguy:
@solomongifford - ah, just noticed the others; I've fixed them in a follow-up commit. Thanks!

PR #80, opened 2016-01-07
Used file module instead of shell command

KrunoKnego:


tasks/setup-Debian.yml
```diff
@@ -22,5 +22,5 @@
   when: mysql_installed.stat.exists == false
 
 - name: Delete innodb log files created by apt package after initial install.
-  shell: "rm -f {{ mysql_datadir }}/ib_logfile[01]"
+  file: path={{ mysql_datadir }}/ib_logfile[01] state=absent
```
geerlingguy: The file module won't take arguments like `[01]` for filename matches, I don't think (that could be why this failed).

geerlingguy:
This has been fixed in a separate issue/PR; sorry about not getting to it sooner!

Issue #77, opened 2015-12-15
Use Docker-based Travis CI tests for multi-platform support


geerlingguy:
See similar: https://github.com/geerlingguy/ansible-role-apache/pull/60

geerlingguy:
Fixed in #84

PR #84, opened 2016-02-03
Issue #77: Add docker based tests

### LINKED ISSUE #55: Better OS/vendor matrix
I saw a very clever solution for supporting multiple vendors:

https://github.com/HanXHX/ansible-mysql

However, it does not support multiple OS Families.

Here is a support matrix that I could imagine:

- Family: Debian, RedHat
- Vendor: MySQL, MariaDB, Percona, Other (?)

For each OS-Vendor pair you could provide a variable list (packages, daemon, etc) and an installation file (Debian/mysql.yml).

This way you could bundle any number of installation profiles out-of-the-box (without the need for overriding packages for mariadb, etc.)

oxyc:
Over at Drupal VM this role is the one that seems to be failing the most in the Docker tests, so I'm adding the setup here too. Currently everything will fail but maybe travis can help us work through all the errors on at a time. I'm getting some weird inconsistencies when running it locally...

Once we get stuff working we can just create a separate `test.yml` for CentOS 7 (or fix #55, which I'd prefer for upstream tests).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,11 @@ env:
     init: /sbin/init
     run_opts: ""
     playbook: test.yml
-  - distribution: ubuntu
-    version: 12.04
-    init: /sbin/init
-    run_opts: ""
-    playbook: test.yml
+  # - distribution: ubuntu
+  #   version: 12.04
+  #   init: /sbin/init
+  #   run_opts: ""
+  #   playbook: test.yml
 
 services:
   - docker
@@ -70,4 +70,4 @@ script:
     || (echo 'MySQL not running' && exit 1)
 
   # Clean up
-  - 'sudo docker stop "$(cat ${container_id})"'
+  - sudo docker stop "$(cat ${container_id})"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,25 +2,26 @@
 sudo: required
 
 env:
-  matrix:
-    - distribution: centos
-      version: 6
-      init: /sbin/init
-      run_opts: ""
-    - distribution: centos
-      version: 7
-      init: /usr/lib/systemd/systemd
-      run_opts: "--privileged --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
-    - distribution: ubuntu
-      version: 14.04
-      init: /sbin/init
-      run_opts: ""
-    - distribution: ubuntu
-      version: 12.04
-      init: /sbin/init
-      run_opts: ""
-  global:
-    - testfile: test.yml
+  - distribution: centos
+    version: 6
+    init: /sbin/init
+    run_opts: ""
+    playbook: test.yml
+  - distribution: centos
+    version: 7
+    init: /usr/lib/systemd/systemd
+    run_opts: "--privileged --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
+    playbook: centos-7-test.yml
+  - distribution: ubuntu
+    version: 14.04
+    init: /sbin/init
+    run_opts: ""
+    playbook: test.yml
+  - distribution: ubuntu
+    version: 12.04
+    init: /sbin/init
+    run_opts: ""
+    playbook: test.yml
 
 services:
   - docker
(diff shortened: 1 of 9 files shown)


oxyc:
Wonder why travis isnt running the builds...

Wrote a quick very ugly python script (never really used python but bash dictionaries suck) for generating a runnable script as travis compile doesnt work with `env`. https://gist.github.com/oxyc/01b757997afd06367959

```sh
wget https://gist.githubusercontent.com/oxyc/01b757997afd06367959/raw/a3f229bde18897fbf8d5a98a59e5b9ad3067859d/create-script.py
python create-script.py >| run.sh

# Verify that it's correct and maybe comment out some build you don't want to run while testing.
vim run.sh

bash run.sh
```

Oh and it only works if you use the YAML syntax for globals (not `SITE=test.yml`) and only when `matrix` is used. Just wrote it for these tests :P

oxyc:
- CentOS 6 passed the idempotence test, but not the two connection tests.
- CentOS 7 with the correct variables passed the idempotence test. Passes both connection tests.
- Both Ubuntu 12.04 and 14.04 fail with the following:

```
TASK [role_under_test : Disallow root login remotely] **************************
failed: [localhost] => (item=DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')) => {"changed": false, "cmd": ["mysql", "-NBe", "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')"], "delta": "0:00:00.059474", "end": "2016-02-03 22:05:07.879763", "failed": true, "item": "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')", "rc": 1, "start": "2016-02-03 22:05:07.820289", "stderr": "ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)", "stdout": "", "stdout_lines": [], "warnings": []}
```

oxyc:
So the ubuntu errors are because `upstart` does not work in Docker. http://stackoverflow.com/a/29640016/319855

Replacing it with systemd is a pain for both Ubuntu 12.04 and 14.04 as well...

oxyc:
Okay now the test run and pass on CentOS 6 and CentOS 7. For now I cant get it working under Ubuntu due to upstart/systemd madness. Good luck @geerlingguy

geerlingguy:
@oxyc - Haha, thanks so much.

I'm thinking we can merge this (after I take a little deeper look), make sure the testing is happy in CentOS, and make it a priority to figure out the best way to do the testing in Ubuntu in Docker. Worst possible case, we could make a frankentest that does RHEL-derivatives inside Docker and then runs the playbook direct in Travis to at least hit 12.04... but that is a non-starter for me.

oxyc:
So `phusion/baseimage` fails in the same spot. Maybe there's something missing hmmm....

LewisW:
I have had this role working inside of Docker for a while. The upstart madness can be worked around with [symlinking upstart](https://github.com/LewisW/docker-test-runner/blob/master/Dockerfile#L76) to [something that just proxies the requests to sysvinit](https://github.com/LewisW/docker-test-runner/blob/master/scripts/initctl_faker). It's a bit ugly, but it works well.

I'll do a blog post explaining a bit more about it and why it's necessary, but for now feel free to ask me any questions. There are probably other workarounds that I've had to do to fix issues, but I can't think of any off the top of my head.

oxyc:
@LewisW absolutely awesome! I'll give that a try later when I have some free time. Thanks for the help.

oxyc:
So we got 14.04 passing thanks to your fix! I'll play around a bit more later with 12.04 and your Dockerfile as a reference @LewisW. I saw you used 14.04 for your own `baseimage-docker`, have you ever tried running it on 12.04?

LewisW:
No sorry! I'd imagine the initctl faker is being called in a way I didn't code for or something simple like that. I'll try have a play this weekend when I do the blog post.

geerlingguy:
@oxyc - Worst case, I'd be willing to switch tests to 14/centos6/7; still better coverage than I have right now with 12 only.

oxyc:
This should now pass for Ubuntu 14.04, Centos 6 and Centos 7. I left Ubuntu 12.04 tests disabled until for now.

PR #87, opened 2016-02-12
Add a variable to permit ansible to manage service states

nerzhul:


paulmaunders:
I also think this would be quite useful to have. 

I am currently working on a playbook to synchronise a slave from a master server, and mysql needs to be stopped until the data folders have been copied. I'd like to use this role to install MySQL but to start it myself later.

nerzhul:
@geerlingguy can you merge this ?

gdelafond:
👍
it could be very useful, but I think there should be two variables : 
- mysql_service_manage :  for default status (started at boot)   
- mysql_service_restart : for the restart notify (which will be only used if mysql_service_manage == True)
@paulmaunders what do you think ?

PR #92, opened 2016-03-15
Add password fileds for blank passwords on CentOS 7 and MariaDB

pecav89:
Found that blank passwords now need to be specified as ''.


geerlingguy:
The password should be automatically picked up from the `~/.my.cnf` file in the root user's account, which is set up by this role as part of the setup process. Passwords shouldn't ever be specified on the command line.

pecav89:
Cool, thanks - I'll try that. :)

Issue #63, opened 2015-10-04
Running role on ubuntu 15.04/16.04 fails


peteraba:
I used the following vagrant box: https://github.com/kraksoft/vagrant-box-ubuntu/releases/download/15.04/ubuntu-15.04-amd64.box Ubuntu 15.04 (based on amd64 server iso file)

$ vagrant up
```
[...]
skipping: [default]
 _______________________________________________________
/ TASK: geerlingguy.mysql | Ensure MySQL is started and \
\ enabled on boot.                                      /
 -------------------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


failed: [default] => {"failed": true}
msg: Job for mysql.service failed. See "systemctl status mysql.service" and "journalctl -xe" for details.


FATAL: all hosts have already failed -- aborting
 ____________
< PLAY RECAP >
 ------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


           to retry, use: --limit @/home/peteraba/site.retry

default                    : ok=18   changed=8    unreachable=0    failed=1   

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
vagrant up  190.06s user 7.51s system 12% cpu 26:17.54 total

```
$ vagrant ssh
```
Welcome to Ubuntu 15.04 (GNU/Linux 3.19.0-15-generic x86_64)

 * Documentation:  https://help.ubuntu.com/
Last login: Sun Oct  4 09:38:18 2015 from 10.0.2.2

```
vagrant@vagrant-ubuntu-trusty:~$ systemctl status mysql.service
```
● mysql.service - MySQL Community Server
   Loaded: loaded (/lib/systemd/system/mysql.service; enabled; vendor preset: enabled)
   Active: activating (start-post) (Result: exit-code) since Sun 2015-10-04 09:48:19 GMT; 3min 22s ago
  Process: 10148 ExecStart=/usr/bin/mysqld_safe (code=exited, status=2)
  Process: 10145 ExecStartPre=/usr/share/mysql/mysql-systemd-start pre (code=exited, status=0/SUCCESS)
 Main PID: 10148 (code=exited, status=2);         : 10149 (mysql-systemd-s)
   CGroup: /system.slice/mysql.service
           └─control
             ├─10149 /bin/bash /usr/share/mysql/mysql-systemd-start post
             └─11170 sleep 1

[...]

PR #93, opened 2016-03-25
Compatibility for mysqld 5.7 - deprecate config values, handle temporary root password

olenm:
add logic for mysqld 5.7:
- use temporary root password on install
- cli or .my.cnf using password 
deprecate variables not used in mysqld 5.7

tested on COS 7, mysqld version 5.7.11 (https://aws.amazon.com/marketplace/pp/B00UVT5VUO)


olenm:
appears the tests need to be configured for a centos 7 with mysqld and one with mariadb

Why it is reporting 'mysqld' does not exist in the path for COS6 or should it be hard-coded to /usr/sbin/mysqld for both centos 6 and 7?
Would an absolute path solve this for ubuntu?

sbrinkerhoff:
Is there any plan to merge this into the role?  I've been using olenm's fork to install MySQL, and would prefer to move back to this as my standard.

geerlingguy:
Yes, at some point; since it's not passing the automated tests, it means I'll need to take some time to dig into it before merging, and right now I'm wrapping up a large project and heading off to a week long vacation next week, so I can guarantee it won't happen within 1-2 weeks... but it will happen (at least, MySQL 5.7 will soon be supported).

geerlingguy:
I'm closing this PR in favor of simply getting 5.7.x full support over in https://github.com/geerlingguy/ansible-role-mysql/issues/63. If you'd like to propose MySQL root password handling changes, please submit a separate PR once #63 is complete.

PR #97, opened 2016-04-03
lets mysql_db defines the default encoding

drzraf:
cf:
- https://docs.ansible.com/ansible/playbooks_filters.html#omitting-undefined-variables-and-parameters
- https://docs.ansible.com/ansible/playbooks_filters.html#omitting-undefined-variables-and-parameters
  Lets machine default apply. They probably are the best of all the possible defaults.

### DIFF
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -2,7 +2,7 @@
 - name: Ensure MySQL databases are present.
   mysql_db:
     name: "{{ item.name }}"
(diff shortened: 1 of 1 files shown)


geerlingguy:
Going to do a slight bit of extra testing on this just to make sure it won't cause much breakage (some people may be relying on the current default). I do like using `omit` instead, and that wasn't an option until 1.8+, so that's the main reason I hadn't been using it in many places in my roles.

PR #99, opened 2016-04-03
don't fail in --check mode

drzraf:
In check-mode, mysql_root_hosts is an object containing `{'skipped': True}`
The when condition apply for *each* item (thus it does not prevent task execution, even if the with_items parameter isn't an array (the `when` will be run `mysql_root_hosts.stdout_lines | count` times).
Thus in --check mode, and to avoid an error, we need to set a valid (an noop) default value)


drzraf:
I don't understand anything at this build failure log.

geerlingguy:
This was fixed in https://github.com/geerlingguy/ansible-role-mysql/pull/130

geerlingguy:
This was fixed in https://github.com/geerlingguy/ansible-role-mysql/pull/130

Issue #95, opened 2016-03-27
mysql_datadir is not respected


abriel:
Hello,

I try to use non standard path for data. I set up `mysql_datadir` , which reflects in configuration,
but data has not moved or created, so start is failed.

Thanks.

geerlingguy:
@abriel - This role is only meant to set up the initial configuration or be adapted to an existing configuration. I don't intend to support things like moving config locations, moving the datadir, etc.—for that, you'll need to write a separate task-specific playbook, as it's a bit more involved, requiring service shutdown, file copies, reconfiguration, then restart.

PR #101, opened 2016-04-06
Support for RHEL 7 mariadb and issue #95

dsglaser:
Added support for RHEL 7 mariadb. Unfortunately it makes a change to tasks/main.yml to support it. 

Fixed issue #95 by making sure that datadir is created and selinux context set on the directory. 

Feel free to use or reject as needed. These were done to support our enterprise installations where I work


tasks/main.yml
```diff
@@ -2,6 +2,17 @@
 # Include variables and define needed variables.
 - name: Include OS-specific variables.
   include_vars: "{{ ansible_os_family }}.yml"
+  when: ansible_os_family  != "RedHat"
+
+# Include RedHat 6 variables (mysql)
+- name: Include OS-specific variables.
+  include_vars: "{{ ansible_os_family }}-{{ ansible_lsb.major_release }}.yml"
```
geerlingguy: If `ansible_lsb.major_release` just outputs `6` or `7`, no need for two tasks, we can just do one task without a `when` condition.

vars/RedHat-7.yml
```diff
@@ -0,0 +1,15 @@
+---
+__mysql_daemon: mariadb
+__mysql_packages:
```
geerlingguy: I like this approach! No need to have people manually override just for RHEL 7.

geerlingguy:
I like this approach, but I have a couple comments for review. Thanks!

tasks/main.yml
dsglaser: Looks good to me.

geerlingguy:
I based https://github.com/geerlingguy/ansible-role-mysql/pull/116 off this PR, and merged. Thanks again!

PR #103, opened 2016-04-15
ansible doesn't have a create-if-not-exists

solomongifford:
We shouldn't be changing the date each time this is run.  While this doesn't make a difference on an installation run, it does make a difference if you are tracking changes.


geerlingguy:
True... see https://github.com/ansible/ansible-modules-core/issues/170

One annoying side-effect—this may toss up a deprecation warning/notice with Ansible 2.x out of the box :/

PR #104, opened 2016-04-15

solomongifford:
This is an alternate to https://github.com/geerlingguy/ansible-role-mysql/pull/13.  It should be backwards compatible.  However, if the user runs as a user other than root, it creates a root user in mysql for that user and deploys the appropriate my.cnf.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,10 +1,14 @@
 ---
 mysql_user_home: /root
+mysql_user_name: root
+mysql_user_password: root
+mysql_root_home: /root
 mysql_root_username: root
 mysql_root_password: root
 
 # Set this to `yes` to forcibly update the root password.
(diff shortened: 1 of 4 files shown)


defaults/main.yml
```diff
@@ -1,10 +1,14 @@
 ---
 mysql_user_home: /root
+mysql_user_name: root
```
geerlingguy: Can you add in a line of docs/comments here explaining the difference between `mysql_user_*` and `mysql_root_*` options; just because it's not obvious, especially since they all default to `/?root` :)

bmsupport: Done.

solomongifford: done

geerlingguy:
Rebased in https://github.com/geerlingguy/ansible-role-mysql/pull/188

PR #105, opened 2016-04-15
support for systemd on Centos7

solomongifford:
Centos 7 uses mariadb/systemd.  This deploys the template - while more work can be done in the future, this gets the support started and allows the systemd file to be overridden at the playbook level.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -42,6 +42,16 @@
     mode: 0644
   when: mysql_slow_query_log_enabled
 
+- name: Set systemd for service
+  template:
+    src: mariadb.service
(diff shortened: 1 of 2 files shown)


geerlingguy:
The default package for mariadb seems to work okay (with it's own systemd setup), at least according to testing on Drupal VM http://docs.drupalvm.com/en/latest/other/base-os/#redhat-enterprise-linux-centos-7

geerlingguy:
Closing this tentatively... do you need this to get MariaDB working? I have tested on CentOS 7 and am completing some testing on Ubuntu 16.04 as well, and it seems MariaDB at least works okay.

PR #114, opened 2016-04-28
Add mysql_skip_name_resolve parameter

nerzhul:
This prevent MySQL DNS name resolution.

Default: false

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -27,6 +27,7 @@ mysql_port: "3306"
 mysql_bind_address: '0.0.0.0'
 mysql_datadir: /var/lib/mysql
 mysql_pid_file: /var/run/mysqld/mysqld.pid
+mysql_skip_name_resolve: no
 
 # Slow query log settings.
 mysql_slow_query_log_enabled: no
--- a/templates/my.cnf.j2
+++ b/templates/my.cnf.j2
@@ -9,6 +9,9 @@ bind-address = {{ mysql_bind_address }}
 datadir = {{ mysql_datadir }}
 socket = {{ mysql_socket }}
 pid-file = {{ mysql_pid_file }}
+{% if mysql_skip_name_resolve %}
+skip-name-resolve
+{% endif %}
 
 # Logging configuration.
 {% if mysql_log_error == 'syslog' or mysql_log == 'syslog' %}


geerlingguy:
I can't find the documentation for the current default—but I'd like to confirm what the MySQL installation default is, and set that inside `defaults.yml`, then allow users to override on their own if they want to.

geerlingguy:
Also, it looks like this option can't be used in the way that is being templated with MySQL < 5.5 (I think CentOS 6 installs 5.1 or 5.2).

nerzhul:
@geerlingguy seems you didn't know mysql very well or didn't look at the documentation. I have some MySQL in production.

[PROD][08:11]root@inf-virt-eta012:~# mysql
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 46415
Server version: 5.1.56-log MySQL Community Server (GPL)

Copyright (c) 2000, 2010, Oracle and/or its affiliates. All rights reserved.
This software comes with ABSOLUTELY NO WARRANTY. This is free software,
and you are welcome to modify and redistribute it under the GPL v2 license

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> show global variables like 'skip%';
+-----------------------+-------+
| Variable_name         | Value |
+-----------------------+-------+
| skip_external_locking | ON    |
| skip_name_resolve     | ON    |
| skip_networking       | OFF   |
| skip_show_database    | OFF   |
+-----------------------+-------+
4 rows in set (0.00 sec)

nerzhul:
Also the current default is 1 for every installation because MySQL by default resolve all hostnames for its acls

nerzhul:
Also, if you search skip_name_resolve into google you will find the official doc for MySQL 5.1 => https://dev.mysql.com/doc/refman/5.1/en/server-options.html#option_mysqld_skip-name-resolve

geerlingguy:
> @geerlingguy seems you didn't know mysql very well or didn't look at the documentation. I have some MySQL 5.1 in production.

@nerzhul - I'm looking through the documentation for 5.5 and 5.7, neither of which are the same as 5.1; 5.1 is included with the system packages in CentOS 6, but I have very few servers that run that system anymore; most servers I manage are on Ubuntu 14.04 (defaults to 5.5), CentOS 7 (defaults to MariaDB 5.5), and I'm moving some to Ubuntu 16.04 (defaults to 5.7)... therefore the configuration from 5.1 in many different ways breaks when being used with more modern MySQL versions.

The 5.5 and 5.7 documentation mentions the option, but doesn't state the default anywhere I could tell... I don't have time right now to spin up a test environment and check the default, but I will at some point.

One aside: I generally haven't hit this issue because most MySQL installations I've ever worked with have one or very few client machines interacting with it (e.g. 1-100 webservers, maybe one or two admins who need to log in from time to time), so the DNS lookup isn't expensive, and isn't noticeable at all after first connection due to the DNS cache.

Disabling the lookup is somewhat an edge case, but as this role _is_ used by some who might consider changing the default, I'm willing to accept it, but I don't want to change the MySQL defaults for 5.5/5.7 lightly.

nerzhul:
@geerlingguy i'm using this role for my 35 MySQL production servers and i need this option to : 
* optimize a little bit the bandwidth
* have a better user IP matching for AWS because AWS reverse IPs are not sexy and a pain to manage using ansible on AWS.

nerzhul:
@geerlingguy please also note the default is to NOT disable, as i set. skip_name_resolve = 0 => resoluve :)

nerzhul:
Oh i see, i set it to 1 in my PR, sorry, ti's a mistake, i will fix it

nerzhul:
This is now fixed and rebase, default is properly "RESOLVED" (skip_name_resolve: 0)
Now you can merge it without problem, sorry for the inconvenience

geerlingguy:
@nerzhul - haha, no problem. I'll check things over and see what's up with the Travis failures.

geerlingguy:
Looks like `/usr/libexec/mysqld: option '--skip-name-resolve' cannot take an argument` is the problem—for MySQL < 5.5, we need to set that option in my.cnf without the `= {{ mysql_skip_name_resolve }}`.

nerzhul:
@geerlingguy i will look at mariadb 10.1 and mySQL 5.7 to be sure we can use a condition instead of a single value

nerzhul:
@geerlingguy you can now merge, i fixed problems with this PR :)

nerzhul:
@geerlingguy ping ?

PR #115, opened 2016-04-28
Fix "when" statement for mysql error log file permissions.

jonpugh:
This is causing failures on my CentOS7 tests for devshop:

https://travis-ci.org/opendevshop/devshop/jobs/126429109#L1705

The error log file is created but permissions are skipped because they are conditional on `mysql_slow_query_log_enabled`, which I assume was a mistake?

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,9 @@ script:
   # Test role.
   - 'sudo docker exec --tty "$(cat ${container_id})" env TERM=xterm ansible-playbook /etc/ansible/roles/role_under_test/tests/${playbook}'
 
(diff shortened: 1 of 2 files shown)


jonpugh:
Doh!

well, at least it only failed on the idempotency test...

```
The command "sudo docker exec "$(cat ${container_id})" ansible-playbook /etc/ansible/roles/role_under_test/tests/${playbook} | grep -q 'changed=0.*failed=0' && (echo 'Idempotence test: pass' && exit 0) || (echo 'Idempotence test: fail' && exit 1)
```

PR #123, opened 2016-05-11
Add mysql tmpdir directive

nerzhul:
Add missing directive to customize tmpdir.
Default is /tmp, link in all mysql installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ The MySQL users and their privileges. A user has the values `name`, `host` (defa
     mysql_port: "3306"
     mysql_bind_address: '0.0.0.0'
     mysql_datadir: /var/lib/mysql
+    mysql_tmpdir: /tmp
 
 Default MySQL connection configuration.
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -28,6 +28,7 @@ mysql_bind_address: '0.0.0.0'
 mysql_datadir: /var/lib/mysql
 mysql_pid_file: /var/run/mysqld/mysqld.pid
 mysql_skip_name_resolve: no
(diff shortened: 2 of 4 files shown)


geerlingguy:
One question—[according to the 5.7 docs](http://dev.mysql.com/doc/refman/5.7/en/temporary-files.html), the temp dir can be any of `/tmp`, `/usr/tmp`, `/var/tmp`, etc... in this case, we're forcing the default `/tmp`. I just want to be sure that at least Debian/Ubuntu/CentOS are not impacted by defaulting to `/tmp` explicitly... but I wonder, what OSes or conditions would not use `/tmp` by default?

> On Unix, MySQL uses the value of the TMPDIR environment variable as the path name of the directory in which to store temporary files. If TMPDIR is not set, MySQL uses the system default, which is usually /tmp, /var/tmp, or /usr/tmp.

nerzhul:
@geerlingguy i have a debian/ubuntu environment, they use /tmp, for centos i don't know the behaviour, it should be tested by testing show global variables like '%tmpdir%';

nerzhul:
Note: i will pull a docker image of centos to know what is the behaviour

nerzhul:
seems it's /tmp too

nerzhul:
@geerlingguy else we can change the value to be False by default and if value set tmpdir to configuration . What do you think about it ?

geerlingguy:
I think we can leave the default `/tmp` then. Just wanted to make sure this wouldn't potentially break on any of the 'officially' supported OSes, and it looks like that's good to go. Can you rebase on master since I merged that other commit?

nerzhul:
@geerlingguy updated !

PR #124, opened 2016-05-12
Permit to remove user, not only create them

nerzhul:


nerzhul:
@geerlingguy a little but useful enhancement to manage users :)

geerlingguy:
Makes perfect sense!

PR #125, opened 2016-05-22
add mysql_binlog_format directive

erkax:
Some application (owncloud in my case) require to use a different binlog_format than the mysql default (statement), so it could be nice to have the opportunity change it directly from your role! 

So I simply added the variable mysql_binlog_format (with a default value of STATEMENT, which is the default for mysql and mariadb)

This change has been tested on CentOS7 and Ubuntu!

Best regards,
Erkax

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ The rest of the settings in `defaults/main.yml` control MySQL's memory usage. Th
 
     mysql_server_id: "1"
     mysql_max_binlog_size: "100M"
-    mysql_binlog_format: "STATEMENT"
+    mysql_binlog_format: "ROW"
     mysql_expire_logs_days: "10"
     mysql_replication_role: ''
     mysql_replication_master: ''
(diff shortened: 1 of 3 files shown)


geerlingguy:
It looks like the default was changed in 5.7.7 to `ROW`:

> Prior to MySQL 5.7.7, statement-based logging format was the default. In MySQL 5.7.7 and later, row-based logging format is the default.

From http://dev.mysql.com/doc/refman/5.7/en/binary-log-setting.html

erkax:
indeed the new default value is ROW, do I have to change the default value for this playbook?

geerlingguy:
It seems like that might be the best default; it works back to 5.5 as well, so it's safe enough for my liking.

erkax:
So now the default value for binlog_format is row, sorry I commited twice... this is my first merge request! Thx for all the code you write and share Jeff

PR #130, opened 2016-06-11
Fix role failing in check mode

UnrealQuester:
If the role is run in check mode, the shell commands never get executed
and the variables used in the next task are undefined, causing the play
to fail.

The commands are now always executed to ensure the variables are
populated.


geerlingguy:
Should be safe, since it's just SELECT statements.

PR #131, opened 2016-06-13
Update secure-installation for correct root password in 5.7+.

geerlingguy:
See: https://github.com/geerlingguy/drupal-vm/issues/650 and http://dev.mysql.com/doc/refman/5.7/en/default-privileges.html

### DIFF
--- a/tasks/secure-installation.yml
+++ b/tasks/secure-installation.yml
@@ -1,9 +1,14 @@
 ---
+- name: Get MySQL version.
+  command: 'mysql --version'
+  register: mysql_cli_version
+  changed_when: false
+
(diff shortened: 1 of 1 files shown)


geerlingguy:
This is a little brittle... but it works for now. I'd like to find a better way to support the two different ways of setting the root password (changes in 5.7.6... why so arbitrary with MySQL's choice to add such a fundamental deployment change in a patch release, no clue :P)

geerlingguy:
See also: https://github.com/ansible/ansible-modules-core/issues/3003 (but that's not currently in an Ansible release, so I still can't rely on Ansible's MySQL modules for this).

PR #132, opened 2016-06-14
Allow RedHat-7 variables to be overridden.

JorisVanEijden:
The file `/vars/RedHat-7.yml` overrides the `mysql_log_error` variable to: "`/var/log/mariadb/mariadb.log`". _This variable can not be overwritten by users of the role._

On my CentOS 7 systems I use the mariaDB packages rather than the CentOS ones so the `/var/log/mariadb` directory does not exist so the `configure.yml` playbook fails when trying to touch that file:
`fatal: [server1]: FAILED! => {"changed": true, "cmd": "touch /var/log/mariadb/mariadb.log", "delta": "0:00:00.004271", "end": "2016-06-14 08:47:03.630648", "failed": true, "rc": 1, "start": "2016-06-14 08:47:03.626377", "stderr": "touch: cannot touch '/var/log/mariadb/mariadb.log': No such file or directory", "stdout": "", "stdout_lines": [], "warnings": ["Consider using file module with state=touch rather than running touch"]}`

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -23,6 +23,21 @@
     mysql_slow_query_log_file: "{{ __mysql_slow_query_log_file }}"
   when: mysql_slow_query_log_file is not defined
 
+- name: Define mysql_log_error.
+  set_fact:
+    mysql_log_error: "{{ __mysql_log_error }}"
+  when: mysql_log_error is not defined
+
+- name: Define mysql_syslog_tag.
+  set_fact:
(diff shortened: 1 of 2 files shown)


geerlingguy:
Hmm... one issue here—since the vars are defined in `defaults.yml`, on CentOS 7 / RHEL 7 machines where you don't have explicitly overridden vars (e.g. as is the case with Drupal VM), then the new `__var` vars won't ever get used, because the vars are already defined via defaults.

So we'd need to find an easier way to let upstream roles override the vars defined via vars includes...

PR #133, opened 2016-06-14
using file module instead of rm

aljazceru:
Using ansible file module instead of rm command

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -22,5 +22,8 @@
   when: mysql_installed.stat.exists == false
 
 - name: Delete innodb log files created by apt package after initial install.
-  shell: "rm -f {{ mysql_datadir }}/ib_logfile[01]"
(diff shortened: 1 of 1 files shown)


geerlingguy:
This makes sense to me, but the main reason I never did this before (I hate that annoying 'best practices' warning too ;) ) is I'm not 100% sure that those are the two and only two files that would need deletion... is there any case where there would be an `ib_logfile2` etc.?

See: http://dba.stackexchange.com/a/27106/3149 — I'm assuming there will only be two, but can anyone confirm? I'm not that deep a DBA to know this definitively :)

geerlingguy:
Docs seem to indicate the only way that number will be more than _1 is if you change the configuration, and as I've never seen otherwise, I think we're safe setting it this way. If someone requests a configurable file list for this someday... that's doable too, but it would probably be an extreme edge case.

Thanks!

PR #135, opened 2016-06-15
Provide overridable OS-specific defaults.

JorisVanEijden:
Makes all the OS-specific vars overridable.


JorisVanEijden:
I've updated the PR to be mergable to master again.

Solves #163 #154

geerlingguy:
This is great! I was actually thinking to myself how annoying the huge list of variable definitions was getting in the main tasks file... and then you go and fix that for me :)

PR #138, opened 2016-07-03
Add a variable to replicate all databases

gdelafond:
Using binlog-do-db is quite complex and dangerous, this new variable allow to create a full server replication.

More information:
- http://dev.mysql.com/doc/refman/5.7/en/replication-rules-db-options.html
- https://www.percona.com/blog/2009/05/14/why-mysqls-binlog-do-db-option-is-dangerous/


geerlingguy:
Will this work back to MySQL 5.5/possibly earlier versions correctly? I haven't gotten 100% up to speed on 5.7+ since I'm only switching some of my servers to that version this year, but I don't want to make a change that could break many LTS users' replication installs.

gdelafond:
Yes, no problem with MySQL 5.5 and earlier. And I set the default to mysql_replicate_all_databases to False, in order to not change the way it was working before.

When mysql_replicate_all_databases is to True in the playbook, it will only bypass the part which adds "binlog_do_db" and "binlog_ignore_db".

zx1986:
@gdelafond Does this commit finish the feature?

gdelafond:
@zx1986 yes

humiaozuzu:
@geerlingguy Can you merge this pr now? We just run into the problem on our production servers described in https://www.percona.com/blog/2009/05/14/why-mysqls-binlog-do-db-option-is-dangerous/
😢

gdelafond:
I will rebase this pull request.

PR #142, opened 2016-07-13
character-set-server = utf8mb4

iainhouston:
Debian MySQL 5.5.49 was warning that 4 byte utf8 was not enabled - utf8mb4 characters set enables this. Suppressing warning only important in our case in that it marked Aegir tasks with warnings.

### DIFF
--- a/templates/my.cnf.j2
+++ b/templates/my.cnf.j2
@@ -2,10 +2,14 @@
 #password = your_password
 port = {{ mysql_port }}
 socket = {{ mysql_socket }}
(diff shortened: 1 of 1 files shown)


geerlingguy:
None of these variables have defaults defined, so they break: `AnsibleUndefinedVariable: Unable to look up a name or access an attribute in template string`

You would need to either add defaults or use `| default('')` or something to that effect.

iainhouston:
Oops. Didn't include all changes in a single PR

PR #144, opened 2016-07-15
default-character-set: utf8mb4

iainhouston:
Apologies for not properly including default vars with template changes in previous PR

### DIFF
--- a/templates/my.cnf.j2
+++ b/templates/my.cnf.j2
@@ -5,7 +5,6 @@ socket = {{ mysql_socket }}
 {% if ansible_os_family == 'Debian' %}
 default-character-set = {{ mysql_default-character-set }}
 {% endif %}
-datadir = {{ mysql_datadir }}
 
 [mysqld]
 port = {{ mysql_port }}
@@ -15,6 +14,7 @@ character-set-client-handshake = {{ mysql_character-set-client-handshake }}
 character-set-server = {{ mysql_default-character-set }}
 collation-server = {{ mysql_collation-server }}
 {% endif %}
+datadir = {{ mysql_datadir }}
 socket = {{ mysql_socket }}
 pid-file = {{ mysql_pid_file }}
 {% if mysql_skip_name_resolve %}
--- a/templates/my.cnf.j2
+++ b/templates/my.cnf.j2
@@ -2,11 +2,19 @@
 #password = your_password
 port = {{ mysql_port }}
 socket = {{ mysql_socket }}
+{% if ansible_os_family == 'Debian' %}
+default-character-set = {{ mysql_default-character-set }}
+{% endif %}
+datadir = {{ mysql_datadir }}
 
 [mysqld]
 port = {{ mysql_port }}
 bind-address = {{ mysql_bind_address }}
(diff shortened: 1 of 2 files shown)


templates/my.cnf.j2
```diff
@@ -2,10 +2,18 @@
 #password = your_password
 port = {{ mysql_port }}
 socket = {{ mysql_socket }}
+{% if ansible_os_family == 'Debian' %}
```
geerlingguy: I'd rather not flag vars based on OS, but rather on MySQL version... is `default-character-set` available in older MySQL versions? If so, we could just leave it without an if...

templates/my.cnf.j2
```diff
@@ -2,10 +2,18 @@
...
+default-character-set = {{ mysql_default-character-set }}
+{% endif %}
 
 [mysqld]
 port = {{ mysql_port }}
 bind-address = {{ mysql_bind_address }}
+{% if ansible_os_family == 'Debian' %}
+character-set-client-handshake = {{ mysql_character-set-client-handshake }}
```
geerlingguy: Same for these—if they work with MySQL 5.1 and/or 5.5 (as well as 5.7), then they don't need to be in an if. Otherwise, we should base on version/supported features rather than OS.

templates/my.cnf.j2
iainhouston: Sure. Will investigate backwards-compatibility. (I should have done that anyway :-()

steffkes:
it's kinda hard to find original mysql 5.1 documentation online - they always redirect to the latest 5.7 release. the only thing i was able to find is a bug report from about six years ago saying:

> The syntax '--default-character-set' is deprecated and will be removed in a future release. Please use --character-set-server instead.
-- https://bugs.mysql.com/bug.php?id=52047

Which for me does imply, that the setting should be good even for very old mysql releases. Other than this, all of the used options are available at least since 5.5: [mysql.default-character-set](http://dev.mysql.com/doc/refman/5.5/en/mysql-command-options.html#option_mysql_default-character-set), [mysqld.character-set-server](http://dev.mysql.com/doc/refman/5.5/en/server-options.html#option_mysqld_character-set-server), [mysqld.collation-server](http://dev.mysql.com/doc/refman/5.5/en/server-options.html#option_mysqld_collation-server)

About [mysqld.character-set-client-handshake](http://dev.mysql.com/doc/refman/5.5/en/server-options.html#option_mysqld_character-set-client-handshake) - i'm not sure how to go with this, since it's a `boolean` appending `true|false` doesn't normally do the trick. also given that

> Do not ignore character set information sent by the client. To ignore client information and use the default server character set, use `--skip-character-set-client-handshake`; this makes MySQL behave like MySQL 4.0.
-- http://dev.mysql.com/doc/refman/5.5/en/server-options.html#option_mysqld_character-set-client-handshake

implies we'd have to use another argument altogether.

And if possible, can we introduce that change w/o a BC-Break? I don't see why we have to provide defaults for those settings. if someone wants to have a global charset-config, he's free to add those to his ansible vars -- otherwise we leave it be like it is right now.

wdyt @iainhouston?

PR #147, opened 2016-07-26
add option for a global/default charset

steffkes:
i know, that the list for `mysql_databases` does allow `encoding` as well as `collation` - but that is only applied on a per-database level.

i wanted my mysql server to apply a global/default charset for everything. didn't have an option for that yet, so i've tried to add one. i did leave out a setting for `defaults/main.yml` on purpose, using `|default(None)` in the templates to check if a value does exist at all.

if you like to be it another way, let me know @geerlingguy


geerlingguy:
Please see #144

steffkes:
Uhm, indeed - i didn't see this, thanks for pointing this out @geerlingguy . i'll have a look and see why the tests are failing and if i can do anything about it

Issue #150, opened 2016-07-31
Remove shell-module touch, use file-module instead. 


rpsu:
Re: https://github.com/geerlingguy/drupal-vm/pull/836#issuecomment-236399095

rpsu:
Actually pull request is in #151

PR #151, opened 2016-07-31
Remove shell-module touch in mysql role, use file-module instead.

rpsu:
#150

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -30,7 +30,9 @@
   notify: restart mysql
 
 - name: Create slow query log file (if configured).
-  shell: "touch {{ mysql_slow_query_log_file }} creates={{ mysql_slow_query_log_file }}"
+  file:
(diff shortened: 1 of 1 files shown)


geerlingguy:
It looks like this breaks idempotence:

```
TASK [role_under_test : Create error log file (if configured).] ****************
changed: [localhost]
```

geerlingguy:
The task would need a `changed_when: false` added (as per https://github.com/ansible/ansible/issues/7490#issuecomment-154117597). See also: https://github.com/ansible/ansible-modules-core/issues/170

jp1987:
Maybe add a warn=no instead then?

geerlingguy:
@jp1987 - Yeah, that could work.

geerlingguy:
Fixed in https://github.com/geerlingguy/ansible-role-mysql/commit/f45e1a892ce13eeadda434d028915b40db0db29a

PR #156, opened 2016-08-17
Auto-create .my.cnf file with default pass, if and only if it is necessary

robmccardle:
Hey Jeff - first, thank you for sharing this awesome playbook.

Your note in the README mentions:

> "Note: If you get an error like ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES) after a failed or interrupted playbook run"

I found that I got that error every time on my system and it was definitely not due to being interrupted as is against stable local VMs:

- Controller / Virtual Box Host: Mac El Capitan 10.11.5 (15F34)
- Target Virtual Box Guest: Ubuntu Server 14.04 64 Bit
- ansible 2.1.1.0

I've automated your solution of creating a temporary .my.cnf with a blank root password (if and only if it doesn't exist) which then gets overwritten later. Making that behaviour part of the playbook seems cleaner and works great for me on initial and repeat test runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The MySQL root user account password.
 
 Whether to force update the MySQL root user's password. By default, this role will only change the root user's password when MySQL is first configured. You can force an update by setting this to `yes`.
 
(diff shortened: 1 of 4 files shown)


robmccardle:
Ah man, so annoying. It works for Ubuntu which is all I need but Travis is still showing that this hangs on Redhat (Centos version 6 & 7) even when I moved the code into setup-Debian.yml and I don't know exactly why. Up to you/others if you want to fix/merge, I'll use my fork. Thanks again

PR #158, opened 2016-08-24
Have option to enable query cache

BanovMiroslav:
By default query_cache_type = 0, which means query cache is OFF.

Adding the parameter to be able to change it, and setting in ON by default.

Wasn't sure if default should be OFF instead?


geerlingguy:
LGTM, thanks!

PR #160, opened 2016-08-28
allow mysql_users to be also defined in a dict

gronke:
The `mysql_users` can alternatively be defined in a dict structure. Here's what the example from the Readme would look like:

    mysql_users:
      example_user:
        host: "%"
        password: similarly-secure-password
        priv: "example_db.*:ALL"

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ The MySQL databases to create. A database has the values `name`, `encoding` (def
 
     mysql_users: []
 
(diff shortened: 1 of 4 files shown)


geerlingguy:
This change seems to be backwards-incompatible—the build is giving the following error:

```
TASK [role_under_test : Ensure MySQL users are present.] ***********************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "with_dict expects a dict"}
```

If this can be done in a way that's backwards compatible, I'd be willing to incorporate it. Otherwise, it's not a big enough enhancement to go in until the next major release.

gronke:
This can be achieved with a Jinja2 filter outside of this role. Ansible is huffy about passing a non-dict structure to `with_dict`, even if it would not be executed because of the condition in `when`.

PR #166, opened 2016-10-04
Added a gotcha note to the README when trying to login from the CLI

jeroenvdgulik:
This took me about an hour of re-provisioning to figure out that it was doing the install correctly, I was just trying to run mysql from the cli as the wrong user

Signed-off-by: Jeroen v.d. Gulik jeroen@isset.nl


geerlingguy:
Makes sense.

PR #169, opened 2016-10-21
Allow sql_mode to be overridden

dabootski:
This SitePoint tutorial explains our exact reasoning for needing to override this setting: https://www.sitepoint.com/quick-tip-how-to-permanently-change-sql-mode-in-mysql/


aavandry:
need this too, please, accept it

gsvitins:
Would be really usefull indeed!

geerlingguy:
Since this is a non-breaking/backwards-compatible change, I've included it in 2.5.1. Thanks!

Issue #71, opened 2015-11-20
MYSQL Users Varibles


upbeta01:
I seem to found some bugs on the mysql role. I have isolated the issue by trying to create 2 db users (1 root-user, 1 db-user).. running both users on the playbook vars gives out an error, running db-user alone gives out an error -- but running root-user alone gives no error. I tried running "mysql" as root and it directs me to the sql-shell which means that  the ".my.cnf" worked properly.

I am running on Ubuntu14 hosted on Digital Ocean Cloud.

---
# PLAYBOOK ERROR

TASK: [ansible-role-mysql | Ensure MySQL users are present.] ******************
failed: [128.199.222.xxx] => (item={'host': 'localhost', 'password': 'testserver-password', 'name': 'testserver-user', 'priv': '_._:ALL'}) => {"failed": true, "item": {"host": "localhost", "name": "testserver-user", "password": "testserver-password", "priv": "_._:ALL"}}
msg: (1045, "Access denied for user 'root'@'localhost' (using password: YES)")

FATAL: all hosts have already failed -- aborting

---
# PLAYBOOK VARS

``` javascript
mysql_root_password: P4$$w0rd
overwrite_global_mycnf: yes
mysql_root_password_update: yes
mysql_databases:
 - name: testserver-db
   encoding: latin1
   collation: latin1_general_ci
mysql_users:
# - name: testserver-user
#   host: "localhost"
#   password: testserver-password
#   priv: "testserver-db.*:ALL"

- name: root
host: "localhost"
password: P4$$w0rd
priv: "*.*:ALL"
```

aleksandervines:
I got the same error message. It seems the root password is set to blank, not what "mysql_root_password" specifies

georgiana-gligor:
:+1: 
This is also occuring in my playbook.
I am trying to provision a Centos6.7 and 

```
TASK: [geerlingguy.mysql | Disallow root login remotely] **********************
# ...
[...]

PR #170, opened 2016-10-23
Connect to MySQL as root user, with password

d4goxn:
I was having a similar problem to what was discussed in #71, and was able to fix it by supplying login credentials to mysql commands.

### DIFF
--- a/tasks/secure-installation.yml
+++ b/tasks/secure-installation.yml
@@ -5,13 +5,13 @@
   changed_when: false
 
 - name: Disallow root login remotely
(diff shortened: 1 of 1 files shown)


geerlingguy:
It's best to avoid passing passwords in the command itself, as it would be preserved in temp files written to disk and/or the bash history (not likely if using SSH). Ansible does have an option not to log the command, but it's still riskier than I like.

The password should be in `.my.cnf` in the /root user's folder, as that's how (until very recently) MySQL's documentation recommends storing auth credentials for use in running `mysql` commands without supplying credentials in the CLI.

PR #174, opened 2016-10-29
Style: increased readability by use of native yaml

rjongejan:
Hello,

Please accept my changes. I only changed the style to conform to the Ansible best practice number 3 as described here: https://www.ansible.com/blog/ansible-best-practices-essentials

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,3 +1,6 @@
 ---
 - name: restart mysql
-  service: "name={{ mysql_daemon }} state=restarted sleep=5"
+  service:
+    name: "{{ mysql_daemon }}"
+    state: restarted
+    sleep: 5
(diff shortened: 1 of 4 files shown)


geerlingguy:
@rjongejan - Thanks for the PR! While some people do go fully with _only_ YAML-literal object syntax for tasks, I've chosen (for this role and most other Ansible work I do) to use that style for any complex tasks, but for tasks where there's only one or two simple arguments, I use Ansible's shorthand `key=value` syntax.

I write a bit about this in a best practices chapter in Ansible for DevOps—it's not as black-and-white as that blog post mentions... the main reason for using the YAML style is for readability and easier diffing—but for very simple tasks, the readability is just as easy on one short line (I would argue easier), and since it's so simple, it's unlikely to ever need a change, therefore the utility of use for VCS diffs is reduced or nonexistent.

As long as the shorthand syntax is supported by Ansible (and not deprecated), I'm going to keep my roles this way.

PR #176, opened 2016-11-07
Fix mysql_replication conditionals for Ansible 2.2

### LINKED ISSUE #152: Breaking mysql_replication change in Ansible 2.2.0+
2 days ago, the mysql_replication module was changed with this commit (https://github.com/ansible/ansible-modules-extras/commit/fc764553260e02bb7b36bf7c601685206d6f93cf).

This changes the getmaster/getslave functionality to not fail, but rather return a boolean Is_Master/Is_Slave. Thus,the slave tasks at the end of replication.yml will never run.

I think the "slave|failed" check should be changed to " not slave.Is_Slave," or something along those lines, but when then check is skipped (like for a master), that field is not present.

donat-b:
Fixes #152

### DIFF
--- a/tasks/replication.yml
+++ b/tasks/replication.yml
@@ -24,7 +24,7 @@
   delegate_to: "{{ mysql_replication_master }}"
   register: master
   when: >
-    slave|failed
+    slave.Is_Slave is defined and not slave.Is_Slave
     and (mysql_replication_role == 'slave')
     and (mysql_replication_master != '')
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
Can you think of any way to make this work in a backwards-compatible way? I don't want to push out this update for 2.2 if it causes anyone running Ansible < 2.2 to break replication... at least not yet.

donat-b:
I guess this should do:
```
((slave.Is_Slave is defined and not slave.Is_Slave) or (slave.Is_Slave is not defined and slave|failed))
and ...
```

geerlingguy:
Added that hopefully-backwards-compatible addition.

PR #180, opened 2016-11-14
default with_items for conditional role includes and check mode

choglundsp:
Default loops to empty lists in order to support conditional role includes and check mode.

Similar to the following issue: https://github.com/ansible/ansible/issues/13274


choglundsp:
Merge conflicts, closing for now.

PR #182, opened 2016-11-20
Add compatibility with ArchLinux

Anthony25:
Hello,
These modifications make your role compatible with Archlinux. It also extends the compatibility as the role does not break if a system does not have its variable file declared. The installation will not be done, but at least the configuration will work


geerlingguy:
LGTM, thanks!

Issue #73, opened 2015-11-30
CentOS 7 MySQL installation.


prasann:
CentOS 7+ has mariadb database engine instead of mysql. I could see the steps to use this playbook to install MariaDB. 

Is it possible to install MySQL in CentOS 7+using this playbook?

geerlingguy:
It is, but you would need to add the repository to install MySQL packages manually, maybe in `pre_tasks`, prior to running this role.

abriel:
also you could set up role variables:
```
    mysql_packages:
    - mariadb
    - mariadb-server
    - mariadb-libs
    - MySQL-python
    - perl-DBD-MySQL
```

pporada-gl:
@prasann 
I have a role whose sole purpose is to run
```
---
- name: Install the MySQL Community Edition repository
  shell: rpm -Uvh https://dev.mysql.com/get/mysql57-community-release-el7-7.noarch.rpm
  args:
    creates: /etc/yum.repos.d/mysql-community.repo
```

Then with the geerlingguy.mysql role, I can specify the following variables
```
mysql_enablerepo: "mysql55-community"
mysql_packages:
  - mysql-community-client
  - mysql-community-server
```

solomongifford:
Centos 7 uses systemd now, which is not managed by this role.  I'm adding a PR.

bbaassssiiee:
Replacing the distribution's mariadb/mysql on RedHat-like systems is asking for dependency-hell. I would stay away from it.

PR #184, opened 2016-11-28
Debian: Mysql 5.6 and higher

olo-dw:
The current mysql package for Debian is 5.5. If you want to install mysql version 5.6 or 5.7 you can use the repo.mysql.com.

I propose this improvement with the readme and the tests.

Btw, I remove an extra services: - docker.

Source: https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/#repo-qg-apt-repo-manual-setup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,12 @@ env:
     playbook: test.yml
   - distro: debian8
     playbook: test.yml
+  - distro: debian8
+    playbook: debian-mysql5.6-test.yml
   - distro: ubuntu1604
     playbook: test.yml
   - distro: ubuntu1404
     playbook: test.yml
-  - distro: debian8
-    init: /lib/systemd/systemd
-    run_opts: "--privileged --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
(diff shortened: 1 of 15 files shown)


alexandermeindl:
This patch looks great. Only think I've concerns is python-mysqldb. This debian package depends on libmysqlclient18, but mysql 5.7 comes with libmysqlclient20. It would be great, if python-mysqldb packages is configurable if to install or not (with __mysql_packages). With this option I can skip  to install python-mysqldb and use pip to get the right MySQL-python version with depends on libmysqlclient20. What do you think about it?

geerlingguy:
This looks great to me, sorry I haven't had time to check into it earlier. Can you rebase since there were some upstream changes to the test structure? I'd also like to make it so you could use the official PPA for Ubuntu, and the yum repos for CentOS (that way no matter what OS, you could either install the system packages or MySQL's official packages).

geerlingguy:
See related for CentOS: https://github.com/geerlingguy/ansible-role-mysql/issues/73

olo-dw:
Rebase done.

Issue #178, opened 2016-11-08
root .my.cnf won't update after mysql_root_password has changed


senorsmile:
Here's the specific scenario that I ran into: 

- I moved a file in group_vars that made them not apply to the server properly
- I ran ansible against the node, which changed the password in /root/.my.cnf from the actual root mysql password to the default 'root' from the role's defaults/main.yml
- I ran ansible again and get a failure at 
  - 
``` 
TASK [geerlingguy.mysql : Disallow root login remotely] ************************
task path: /var/lib/jenkins/workspace/ansible-docker/roles/geerlingguy.mysql/tasks/secure-installation.yml:2
failed: [icinga-server] (item=DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')) => {"changed": false, "cmd": ["mysql", "-NBe", "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')"], "delta": "0:00:00.004201", "end": "2016-11-08 22:06:31.254759", "failed": true, "item": "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')", "rc": 1, "start": "2016-11-08 22:06:31.250558", "stderr": "ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)", "stdout": "", "stdout_lines": [], "warnings": []} 
```
- I saw the problem and then moved the yml file to the proper place in group_vars
- I verified with a debug message that the mysql_root_password was properly set now
- Running ansible again never gets to the code to update the .my.cnf, because that task is AFTER the above task that errs out.  

Since the current order of things has to do with idempotence as stated here:
```yaml
# Has to be after the root password assignment, for idempotency.
- name: Copy .my.cnf file with root password credentials.
  template:
    src: "user-my.cnf.j2"
    dest: "{{ mysql_user_home }}/.my.cnf"
    owner: root
    group: root
    mode: 0600
```
it's unclear what  a good solution would be for this scenario.  

I was able to get around it by copying that task, putting a temporary copy of the above task before the root .my.cnf task, and later removed it.  This allowed everything to run as expected.

guimaluf:
I'm also experiencing this error. 
How I'm using this role
```
- name: Install MySQL
  include_role:
    name: "geerlingguy.mysql"
  vars:
[...]

PR #185, opened 2016-11-29
Allow the root password to be changed

ccjjmartin:
## Purpose:
- BUG FIX: On machines with passwords that had not been setup yet (no password) the following task would fail:

TASK [ansible-role-mysql : Disallow root login remotely] ***********************
failed: [XXX.XXX.XXX.XXX] (item=DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')) => {"changed": false, "cmd": ["mysql", "-NBe", "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')"], "delta": "0:00:00.004203", "end": "2016-11-29 07:08:49.255720", "failed": true, "item": "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')", "rc": 1, "start": "2016-11-29 07:08:49.251517", "stderr": "ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)", "stdout": "", "stdout_lines": [], "warnings": []}

## Fixes:
- #178 

## Playbook.yml:
```
- hosts: db
  become: yes
  vars_files:
    - main.yml
  roles:
    - ansible-role-mysql
```

## Main.yml:
```
mysql_root_password: ""
mysql_root_password_update: yes

# This section should be unnecessary but I had it in while testing:
mysql_databases:
  - name: drupal
    encoding: latin1
    collation: latin1_general_ci

# This section should be unnecessary but I had it in while testing:
mysql_packages:
  - mariadb-client
  - mariadb-server
  - python-mysqldb
```

## Notes:
- It appears that the default variables in main.yml were not including the user-my.cnf.j2 template by default which wasn't setting the root password
- The mysql server also wasn't being restarted after the new configuration files were dropped into place

## Testing:
- Setup a host called db in (/etc/ansible/hosts # default location) or (/usr/local/etc/ansible/hosts # homebrew installs)
```
[db]
XXX.XXX.XXX.XXX # Your machine ip here
```
- Create the playbook.yml and main.yml files
- Create a roles directory
- Drop this role into the directory
- Run `ansible-playbook playbook.yml` from the project root

## Thanks:
- I bought the Ansible for DevOps book and read it which allowed me to contribute back to this project.  Thanks for writing it!

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -83,7 +83,8 @@ mysql_log: ""
 # mysql_log_error: /var/log/mysql.err
 # mysql_syslog_tag: mysql
 
-mysql_config_include_files: []
+mysql_config_include_files:
+- {src: user-my.cnf.j2, force: true}
 #  - src: path/relative/to/playbook/file.cnf
 #  - { src: path/relative/to/playbook/anotherfile.cnf, force: yes }
 
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -21,13 +21,16 @@
 - name: Copy my.cnf override files into include directory.
   template:
     src: "{{ item.src }}"
-    dest: "{{ mysql_config_include_dir }}/{{ item.src | basename }}"
+    dest: "/root/.my.cnf"
     owner: root
     group: root
     mode: 0644
     force: "{{ item.force | default(False) }}"
   with_items: "{{ mysql_config_include_files }}"
   notify: restart mysql
+
+- name: mysql-restart
+  service: name=mysql state=restarted
 
 - name: Create slow query log file (if configured).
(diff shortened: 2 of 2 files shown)


geerlingguy:
The purpose of the `mysql_config_include_files` variable/functionality is to allow extra configuration files to be added for MySQL for the server itself.

The .my.cnf file is controlled in the `Copy .my.cnf file with root password credentials.` task in `secure-installation.yml`. This PR's approach isn't touching the right area of the role, so please open a separate PR if you can find a way to fix the bug and keep tests passing.

PR #187, opened 2016-12-15
Add query_cache_type configuration

mamchenkov:
`query_cache_type` was missing from the configuration variables.  Without it, the other query cache settings are not very useful, as the default type is 0 (disabled).  This PR adds `mysql_query_cache_type` variable and sets the default to 1, as per http://dev.mysql.com/doc/refman/5.7/en/query-cache-configuration.html


mamchenkov:
Oops ... this is a duplicate of PR #158 .  Sorry.

PR #188, opened 2016-12-16

geerlingguy:
Following up to https://github.com/geerlingguy/ansible-role-mysql/pull/104


geerlingguy:
@solomongifford - Can you take a look and make sure the rebase looks good? I'm okay with this, but will need to do some more testing with Drupal VM and a couple of my other local environments before merging, since it affects setup a little.

Issue #179, opened 2016-11-10
always_run is deprecated


030:
```
[DEPRECATION WARNING]: always_run is deprecated. Use check_mode = no instead..
This feature will be removed in version 2.4. Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.
```

geerlingguy:
The problem is if I switch to `check_mode == no`, then I'll end up requiring everyone update to Ansible 2.2 (because that variable doesn't work in < 2.2). So I'm going to close this and hold off on updating my roles until sometime around when 2.3 comes out.

PR #190, opened 2017-01-03
Replace deprecated always_run argument

anavarre:
With version `2.5.0` I'm getting the following error:

```
[DEPRECATION WARNING]: always_run is deprecated. Use check_mode = no instead..
This feature will be removed in version 2.4. Deprecation warnings
can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

So it looks like a straightforward replacement.


geerlingguy:
Thanks for the PR, but for now, I am choosing to leave this as-is, despite the highly-annoying warnings. See: https://github.com/geerlingguy/ansible-role-mysql/issues/179

PR #197, opened 2017-01-18
Fix for "always_run" deprecation warning

icamys:
In ansible v2.2.1.0 getting deprecation warning:

```
[DEPRECATION WARNING]: always_run is deprecated. Use check_mode = no instead..
This feature will be removed in version 2.4. Deprecation warnings can be disabled
 by setting deprecation_warnings=False in ansible.cfg
```

This commit fixes it by replacing `always_run` with `check_mode`.


geerlingguy:
Please see https://github.com/geerlingguy/ansible-role-mysql/issues/179#issuecomment-259762499 :(

PR #199, opened 2017-02-05
Add the event_scheduler option to my.cnf.j2 with the default value of 'OFF'

TopShelfBullard:
I added the `event_scheduler` option, because we had need of it in a project. Here are the docs for that option: https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_event_scheduler. I thought it could potentially be useful to others.


geerlingguy:
LGTM!

PR #203, opened 2017-02-26
solve "There are problems and -y was used without --force-yes" issue

Aelshafei:
This issue faced in Ubuntu 12.04

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -8,10 +8,10 @@
   when: mysql_installed.stat.exists == false
 
 - name: Ensure MySQL Python libraries are installed.
-  apt: "name=python-mysqldb state=installed"
+  apt: "name=python-mysqldb state=installed force=yes"
 
 - name: Ensure MySQL packages are installed.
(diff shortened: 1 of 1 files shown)


geerlingguy:
@Aelshafei - Interesting, I've never run into this error. Is that happening on a machine where some other version of MySQL was already installed?

geerlingguy:
I'm going to close this as it seems like something I wouldn't use too lightly, and only as a last resort:

> **Force yes**. This is a dangerous option that will cause apt-get to continue without prompting if it is doing something potentially harmful. It should not be used except in very special situations. Using --force-yes can potentially destroy your system!

(from the [apt-get man page](https://linux.die.net/man/8/apt-get)).

In my case, since I usually either build a server new/fresh, or rebuild a server when making any large changes, I would rather figure out what errors were preventing apt-get from automatically installing the package based on my apt priorities. If there's an error, it usually indicates you have priorities misaligned, are using packages from an apt source or PPA that's not in sync with some other dependencies on your system, etc.

PR #208, opened 2017-03-31
Gh 206 207 two extra parameters

030:


defaults/main.yml
```diff
@@ -64,6 +64,8 @@ mysql_query_cache_limit: "1M"
 mysql_max_connections: "151"
 mysql_tmp_table_size: "16M"
 mysql_max_heap_table_size: "16M"
+mysql_group_concat_max_len: 1024
```
surudoiryu: number between quotes like the other variables to keep it consistent.

PR #209, opened 2017-04-01
Don't try to change existing MySQL

nome:
On Debian, we did check whether MySQL is already installed, but we tried to
install mysql_packages in any case. This fails if a non-default MySQL version
is already installed, e.g. mysql-server-5.6 on Ubuntu Trusty.

Therefore, skip package installation step if /etc/init.d/mysql exists.

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -14,6 +14,7 @@
   apt: "name={{ item }} state=installed"
   with_items: "{{ mysql_packages }}"
   register: deb_mysql_install_packages
+  when: mysql_installed.stat.exists == false
 
 # Because Ubuntu starts MySQL as part of the install process, we need to stop
 # mysql and remove the logfiles in case the user set a custom log file size.


geerlingguy:
@nome - In that case, I would make sure that you set `mysql_packages` to a value that installs only the versions/names that you need. E.g.:

```yaml
mysql_packages:
  - mysql-server-5.6
```

That way an addition when condition isn't required, and you can rely on Ansible to know whether or not the package in question is installed.

I'd rather recommend you be precise in configuring what package(s) should be installed than adding this check, which, in some cases, could cause further confusion (e.g. if MySQL had been previously uninstalled but not purged).

nome:
Hm. Point about uninstalled but not purged packages taken. Explicitly setting the package names makes it difficult to write generic playbooks though (specifically, for test cases running in different environments), since I have to know which packages will work under different circumstances (OS, release version, preinstalled packages, ...). :-/

geerlingguy:
@nome - I'd recommend, in that case, making sure your base images from which your servers are built don't have MySQL preinstalled; or if you're using servers that may already have MySQL installed, if you want to use this role with all it's defaults, add a `pre_tasks` task that ensures that other packages are uninstalled, or sets `mysql_packages` to `[]`.

nome:
@geerlingguy Thank you for the suggestions. I think I've finally found a workable solution (which does involve setting `mysql_packages` to `[]`).

PR #224, opened 2017-07-03
Correcting ERROR 1045 password:NO error

sonance207:
Adding root password to each MySQL command to correct ["ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)"] issue in task Update secure-installation.yml

### DIFF
--- a/tasks/secure-installation.yml
+++ b/tasks/secure-installation.yml
@@ -23,9 +23,9 @@
   when: mysql_user_name != mysql_root_username and (mysql_install_packages | bool or mysql_user_password_update)
 
 - name: Disallow root login remotely
-  command: 'mysql -NBe "{{ item }}"'
+  command: 'mysql -NBe "{{ item }}" --password="{{ mysql_root_password }}"'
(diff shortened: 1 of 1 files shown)


sonance207:
@geerlingguy    Hey Jeff, I having the same issue as #192 and I am trying to make it pass the Travis CI build but I am having difficulty finding the role_under_test that is mentioned in the test.yml file.

geerlingguy:
@sonance207 - the `role_under_test` inside the Docker container / test playbook just refers to the role's directory (this project's root), mounted as a volume inside the container.

Usually having to pass the password on the command line just means the `.my.cnf` file is not working correctly, visible to the user running the role, or is not created at all yet. I'd rather avoid having the password exposed as a cli command if at all possible.

Also, unless someone can build a reproducible test failure scenario for this, I don't want to merge any code that would make the process _more_ complex.

sonance207:
Thanks for adding clarity to that, I am closing this issue

PR #226, opened 2017-07-06
fixes #225 include ansible_managed in templates.

### LINKED ISSUE #225: Add ansible_managed to templates
Templates should include the {{ ansible_managed }} info.

rwese:
Modified   templates/my.cnf.j2
Modified   templates/root-my.cnf.j2
Modified   templates/user-my.cnf.j2

### DIFF
--- a/templates/my.cnf.j2
+++ b/templates/my.cnf.j2
@@ -1,3 +1,5 @@
+# {{ ansible_managed }}
+
 [client]
 #password = your_password
 port = {{ mysql_port }}
--- a/templates/root-my.cnf.j2
+++ b/templates/root-my.cnf.j2
@@ -1,3 +1,5 @@
+# {{ ansible_managed }}
+
 [client]
 user="{{ mysql_root_username }}"
 password="{{ mysql_root_password }}"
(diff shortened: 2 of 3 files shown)


templates/my.cnf.j2
```diff
@@ -1,3 +1,7 @@
+
+# Don't edit manually!
+# {{ ansible_managed }}
```
geerlingguy: The `{{ ansible_managed }}` line should suffice. Can you change the files to just have two new lines—first:

```
# {{ ansible_managed }}
```

Then a blank newline before the start of the rest of the file? That follows the convention I use in many of my other roles (e.g. https://github.com/geerlingguy/ansible-role-redis/blob/master/templates/redis.conf.j2 or https://github.com/geerlingguy/ansible-role-ntp/blob/master/templates/ntp.conf.j2).

rwese:
Updated. 

You are right of course that was redundant.

geerlingguy:
Looks good now!

geerlingguy:
I merged, but will wait to release until I have a couple other changes ready, just because this will cause everyone who uses the role to see a `changed` in their playbooks since Ansible will be adding the line in (even if they didn't change any other configs). If I'm going to make a slightly-breaking-ish-change like that, I like to group a few of these changes in for a minor release.

geerlingguy:
Thanks again!

PR #227, opened 2017-07-10
Allow empty sql_mode

popstas:
Some old CMS requires sql_mode = ''.
I changed default `mysql_sql_mode = ''` to `mysql_sql_mode = false`.

It can break backward compatibility if `mysql_sql_mode = ''` stored in group_vars or somewhere else `default/main.yml`.

Possible better way: add something like `mysql_sql_mode = "empty"` and add to template:
```
{% if mysql_sql_mode == 'empty' %}
sql_mode = ''
{% endif %}
```

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -39,7 +39,7 @@ mysql_port: "3306"
 mysql_bind_address: '0.0.0.0'
 mysql_skip_name_resolve: no
 mysql_datadir: /var/lib/mysql
-mysql_sql_mode: ''
+mysql_sql_mode: false
 # The following variables have a default value depending on operating system.
 # mysql_pid_file: /var/run/mysqld/mysqld.pid
 # mysql_socket: /var/lib/mysql/mysql.sock
--- a/templates/my.cnf.j2
+++ b/templates/my.cnf.j2
@@ -12,7 +12,7 @@ pid-file = {{ mysql_pid_file }}
 {% if mysql_skip_name_resolve %}
 skip-name-resolve
 {% endif %}
-{% if mysql_sql_mode %}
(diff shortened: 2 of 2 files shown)


geerlingguy:
@popstas - What CMS does this? I haven't run into issues with some of the more mainstream ones, and I don't want to add an additional condition to the template (increasing complexity for what that variable does) unless it's something that's fairly widespread.

Otherwise, if there's a way to keep it using one condition and defaulting to not printing it (as it stands currently), then I'd be okay merging that.

popstas:
Bitrix CMS and Bitrix24 CRM - https://www.bitrix24.com
It places at 3-rd place in Russia, after Wordpress and Joomla.

I changed sql_mode from '' to false in `default/main.yml` and reduced conditions to one, please check.

juliocanares:
I think this is needed. I tried to pass `mysql_sql_mode: ''`, but is not working it just writes the default value 
```sql
ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
```

juliocanares:
the default value in `defaults/main.yml` is set to `''` but the jinja parse won't do anything

image

popstas:
@juliocanares, you are right, in my role I using
``` yaml
{% if mysql_sql_mode %}
sql_mode = {{ mysql_sql_mode }}
{% endif %}
{% if mysql_sql_mode == '' %}
sql_mode = ''
{% endif %}
```

PR #229, opened 2017-08-10
Create pid folder and persist across reboots

fastest963:
The `/var/run` folder doesn't exist by default and if it doesn't then the pid can't be created and it crashes. Additionally, `/var/run` isn't persisted across reboots and so if the box is rebooted then it starts crashing again.


michael-stevens:
I'm seeing this problem any chance a fix for this could be merged?

PR #234, opened 2017-09-07
Use {{ ansible_managed | comment }} on templates

Skardian:
This pull request fixes the problem we were having with our ansible_managed, as we have a multiline ansible_managed and only the first line gets commented.


keimond:
👍

PR #235, opened 2017-09-11
Add a variable to permit replication when mysql service listens on a different network than ssh service

atbore-phx:
In most production cases and for security reason mysql service is usually binded on a different network than ssh. 
Prior this PR, ansible via the delegate_to under replication task can't reach the server or the slave uses a wrong master address.

This PR adds a variable to permit replication in those cases.

The default behavior is unchanged because under defaults/main.yml:

    mysql_master_hostname: "{{ mysql_replication_master }}"
  
but now mysql_master_hostname can be changed at playbook level.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -120,6 +120,9 @@ mysql_max_binlog_size: "100M"
 mysql_binlog_format: "ROW"
 mysql_expire_logs_days: "10"
 mysql_replication_role: ''
+# the mysql master address reachable from the slave.
 mysql_replication_master: ''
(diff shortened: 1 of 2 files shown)


atbore-phx:
Hi @geerlingguy,
Do you have some time for review ?

PrestanceDesign:
Hi @geerlingguy, 

Being confronted with the same blocking, it would be really useful to merge this pull request for a next version. Unless there is another way?

Cheers.

atbore-phx:
hi @geerlingguy,
it isn't stale, the PR needs review.

PR #236, opened 2017-09-15
Set the min/max word characters in search indexes

bhean:
Hi!

This is a very small modification to allow to set a pair of "fulltext" options: min and max characters in a indexed word (https://dev.mysql.com/doc/refman/5.7/en/fulltext-fine-tuning.html)

This Ansible-role is very nice, and with this small modification it will be very usefull for me!

Thank you!

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -85,6 +85,10 @@ mysql_innodb_log_buffer_size: "8M"
 mysql_innodb_flush_log_at_trx_commit: "1"
 mysql_innodb_lock_wait_timeout: "50"
 
+# These settings require MySQL > 5.6.
+#mysql_innodb_ft_min_token_size: "3"
+#mysql_innodb_ft_max_token_size: "84"
+
 # These settings require MySQL > 5.5.
 mysql_innodb_large_prefix: "1"
 mysql_innodb_file_format: "barracuda"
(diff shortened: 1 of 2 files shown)


bhean:
Now the values are for MyISAM tables... 
In the InnoDB tables you need MySQL >= 5.6, so I removed these settings by the moment

bhean:
@geerlingguy, I opened this PR 2 years ago... please, can you take a look? I think its a good improve, IMHO...

geerlingguy:
These options don't fit what I'd consider the 80% use case for this role and are not used in many installations. In this case, I'd suggest copying the my.cnf.j2 template into your playbook and customizing it as you see fit, and managing it via your playbook (setting `overwrite_global_mycnf: true` to `false` so this role doesn't change that file).

bhean:
These optiones are completely optional... you can use if you need it, and you can NOT if you don't need it (like in the 80% use case that you mention) because the default values setted in the config are the default values of the official MySQL installation. I think this PR extends the features of the role without any "breaking change" and without increase complexity... but its ok. Thanks.

geerlingguy:
@bhean - I understand, but every additional variable added is another variable which _should be_ (but often isn't) tested independently, and which (and this has happened in the past 5 years of this role's maintenance) could be deprecated or renamed in a future MySQL version, which would mean that variable would need to be redone in a backwards-compatible shim.

It's easier to take a 'less is more' approach in this role, and allow users to override the configuration on their own for more esoteric variables (IMO).

PR #240, opened 2018-01-03
Allow databases to be deleted

socketwench:
Right now the state for databases is hard-coded to present. It would be useful to allow databases to be deleted from within the role.


geerlingguy:
Most definitely!

socketwench:
Yay!

PR #245, opened 2018-02-11
Replication sample

vitvegl:
Replication sample was added

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,27 @@ None.
         host: "%"
         password: similarly-secure-password
         priv: "example_db.*:ALL"
+        
+## Replication sample
+```yaml
+  - hosts: master
+    roles:
+      - role: ansible-role-mysql
+    vars:
+      mysql_server_id: 1
+      mysql_replication_role: master
+      mysql_replication_master: master.test
+      mysql_replication_user: {name: repl, host: "%", password: replica, priv: "*.*:REPLICATION SLAVE,REPLICATION CLIENT"}
+
+  - hosts: slave
+    roles:
+      - role: ansible-role-mysql
+    vars:
+      mysql_server_id: 2
+      mysql_replication_role: slave
+      mysql_replication_master: master.test
+      mysql_replication_user: {name: repl, host: master.test, password: replica}
+```
 
 ## License
 
--- a/tasks/replication.yml
+++ b/tasks/replication.yml
@@ -14,6 +14,7 @@
 - name: Check slave replication status.
   mysql_replication:
     mode: getslave
+    login_host: "{{ mysql_replication_user.host }}"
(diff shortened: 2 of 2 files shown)


Maelstromeous:
@geerlingguy Can we please merge this? Been faffing with replication setup all day and the upstream readme isn't overly enlightening :(

README.md
```diff
@@ -182,6 +182,27 @@ None.
...
+      mysql_replication_user: {name: repl, host: "%", password: replica, priv: "*.*:REPLICATION SLAVE,REPLICATION CLIENT"}
+
+  - hosts: slave
+    roles:
+      - role: ansible-role-mysql
+    vars:
+      mysql_server_id: 2
+      mysql_replication_role: slave
+      mysql_replication_master: master.test
+      mysql_replication_user: {name: repl, host: master.test, password: replica}
```
Maelstromeous: `host: master.test` is so confusing. Maybe you should add a comment or otherwise make it obvious it means the IP of the master?

geerlingguy: It makes some sense; it can be a DNS name, and if you're building a local test infra, `.test` is the safest top level TLD to use.

geerlingguy:
As a general note—I prefer to run one play over all the hosts, and set the master / slave details in inventory, so maybe an example more like:

## Master-Slave replication example

### Inventory

```
[db]
master.test mysql_replication_role=master
slave.test mysql_replication_role=slave

[db:vars]
mysql_replication_master=master.test
```

### Playbook

```yaml
---
- hosts: db
  become: yes

  vars:
    mysql_replication_user: {name: 'replication', password: 'secret'}

  roles:
    - geerlingguy.mysql
```

That _should_ work... I'm kind of pulling from one of my other examples in https://github.com/geerlingguy/ansible-for-devops/tree/master/lamp-infrastructure

omissis:
Hi, I tested the change and I can confirm that adding `login_host` to the `Check slave replication status.` task in `replication.yml` does indeed solve a failure I see in vagrant and aws playbooks.
About the docs(sorry for the unprompted opinion :D) > I'd go for what @Maelstromeous proposed rather than having the vars in the inventory, as I prefer not to clutter the inventory file. But again that's just an opinion: perhaps both ways could be presented in the docs so that the users could pick the most appropriate in their situation?

amanhigh:
Can we merge this PR as the replicaton.yml bug is causing check slave status to fail and hence ansible glaxy package is not working.

Mil0dV:
@geerlingguy 's suggestion seems pretty elegant, however, it seems not to work for us due to both nodes getting mysql_server_id 1 and mysql_replication_master not being specified.

caesarleong:
Can we have this PR work?

PR #258, opened 2018-05-16
Allow host of default user to be overridden

trevorr:
Particularly, I need to be able to allow the default user to have a wildcard host (`%`).

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -3,6 +3,7 @@
 # or sudo access
 mysql_user_home: /root
 mysql_user_name: root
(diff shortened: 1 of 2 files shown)


trevorr:
@geerlingguy Why confused? I use this role for deployment of non-production MySQL servers, and I'm using the default user to represent the remote administrator of those servers. In production, we use AWS RDS, so I treat the default user as analogous to its master user. Since one cannot SSH into an RDS instance, our deployment automation assumes that logging into the MySQL host (and then connecting from localhost) is not an option.

PR #261, opened 2018-06-07
Support empty sql-mode

zanardigit:
The mysql parameter *sql-mode* accepts empty value. So, the role variable mysql_sql_mode should accept it as well.


jschlier:
This would be immensly helpful.
While an empty SQL mode is not necessarily desirable, it is sometimes needed for some frameworks to work with the database.

PR #269, opened 2018-08-13
Add support for innodb_default_row_format

litewhatever:
As noted in subject add support for innodb_default_row_format and set default to "dynamic"


toxsick:
Having the same problem, can we get this one in?

PR #273, opened 2018-08-31
Make the role compatible with MySQL 8

andreykaipov:
This role works great, but some configuration doesn't work for MySQL 8. There's a few variables that aren't supported anymore. See https://dev.mysql.com/doc/refman/8.0/en/added-deprecated-removed.html#optvars-removed. I've wrapped those settings around a conditional based on the `mysql_cli_version` fact.

Also I've edited the two tasks that set the root password based on the version so that 8.0 uses the right one.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -1,4 +1,9 @@
 ---
+- name: Get MySQL version.
+  command: 'mysql --version'
+  register: mysql_cli_version
+  changed_when: false
+
 - name: Copy my.cnf global MySQL configuration.
   template:
     src: my.cnf.j2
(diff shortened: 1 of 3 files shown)


geerlingguy:
Thanks!

andreykaipov:
Hi @kishorevavilapalli - I'm not sure. I don't think I tested any kind of upgrade path when I added this. I imagine if [the existing server is prepped properly](https://dev.mysql.com/doc/refman/8.0/en/upgrade-prerequisites.html), then the role would work as expected, but please don't take my word as I'm not a MySQL expert nor have I used this role since this PR.

PR #276, opened 2018-09-11
Option to omit pid-file

bngsudheer:
Having pid-file can cause issues in recent versions of MariaDB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The formats of these are the same as in the `mysql_user` module.
 
 For some recent versions of MariaDB you might not want to specify `pid-file`. For such situations, set a blank value.
 
-    mysql_pid_file: 
+    mysql_pid_file: ''
 
 Default MySQL connection configuration.
 
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Whether MySQL should be enabled on startup.
 
     mysql_config_file: *default value depends on OS*
     mysql_config_include_dir: *default value depends on OS*
-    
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -90,6 +90,10 @@ The formats of these are the same as in the `mysql_user` module.
     mysql_socket: *default value depends on OS*
     mysql_pid_file: *default value depends on OS*
 
+For some recent versions of MariaDB you might not want to specify `pid-file`. For such situations, set a blank value.
+
+    mysql_pid_file:
```
geerlingguy: Let's set it to a blank string to be a little more explicit:

```
    mysql_pid_file: ''
```

README.md
```diff
@@ -90,6 +90,10 @@ The formats of these are the same as in the `mysql_user` module.
     mysql_socket: *default value depends on OS*
     mysql_pid_file: *default value depends on OS*
 
+For some recent versions of MariaDB you might not want to specify `pid-file`. For such situations, set a blank value.
```
geerlingguy: To further connect this line with the variable below, please do `set a blank value:` (instead of the period).

tests/centos-7-mariadb-10-test.yml
```diff
@@ -0,0 +1,23 @@
+---
```
geerlingguy: This looks good, but can you add an automated test to the `.travis.yml` config as well that uses this playbook so it can be covered for the future?

PR #278, opened 2018-09-14
Fix replication

roedie:
The master.File and master.Position seem not to be needed.


geerlingguy:
They might be needed in certain circumstances (and see module docs here): https://docs.ansible.com/ansible/2.5/modules/mysql_replication_module.html

Please point to a use case that's breaking or steps to reproduce if the way this is currently set up is not working somehow.

Thanks!

PR #281, opened 2018-09-15
Added clarity on what mysql_replication_master does and caveats

Maelstromeous:


geerlingguy:
Makes sense to note this in the README. Thanks!

PR #282, opened 2018-09-17
Setup replication before users/databases

roedie:
When a new install is done the databases and users are created before 
the replication is configured. This way the users an databases are not 
replicted.

This commit moves the replication before user/database creation.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -21,6 +21,6 @@
 # Configure MySQL.
 - include_tasks: configure.yml
 - include_tasks: secure-installation.yml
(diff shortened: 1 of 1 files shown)


roedie:
1 check failed, but looking at it it seems like a problem at Travis and not with the PR

roedie:
@geerlingguy Thanks, al checks have succeeded now.

roedie:
@geerlingguy Hi, I just had to use this role again and notices this hasn't been merged. Is it still possible to merge it?

PR #285, opened 2018-09-17
Allow modifying the log_bin settings

roedie:
This commit allow setting the location and the name of the mysql-bin and 
mysql-bin.index.

This is useful for some installations where you want to have the 
database and logfiles on different media.

If not set, the old defaults are used.


roedie:
@geerlingguy Hi, is it possible to merge this one? Or do I need to change things?

PR #299, opened 2018-11-15
Allows sql mode to be empty

nkakouros:
According to the [documentation](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_sql_mode), the default for `sql_mode` is not to be empty. If someone wants to set it to be empty, it is impossible with this role. This PR changes that so that it becomes possible.


tterranigma:
It is just an arbitrary but semantic string to not have `sql_mode` set to the empty string. I chose it as middle ground as the default for this variable is both complex and different per platform.

nkakouros:
Maintaining my own fork.

PR #303, opened 2018-12-13
Support for importing dabase dump with "state: import"

mkotsbak:


mkotsbak:
Please look into the PR and consider merging.

koalo:
This is a trivial change and would help a lot! Any objections against merging this @geerlingguy?

tasks/databases.yml
```diff
@@ -5,4 +5,5 @@
     collation: "{{ item.collation | default('utf8_general_ci') }}"
     encoding: "{{ item.encoding | default('utf8') }}"
     state: "{{ item.state | default('present') }}"
+    target: "{{ item.state | default('') }}"
```
geerlingguy, line 8: If this just used `item.state`, then it wouldn't get a target dump file. This should be a different setting, like `item.target`, and it should `default(omit)`.

PR #304, opened 2018-12-26
Percona/mysql 5.7 on CentOS 7

wouterdt:
Added the possibility to handle the installation of a Mysql/Percona Server 5.7 on CentOS 7.
On Percona & Mysql on Centos the default password handling is completely different than on the MariaDB  (https://www.percona.com/blog/2016/05/18/where-is-the-mysql-5-7-root-password/) This required some splitting and  moving around in secure_installation.yml.
Added a molecule scenario to test setting up the Percona server.
This version still would require you to set the mysql_packages as I did not add the Percona yum repositories to the role.


LBRDan:
Hello there,
is there a chance this is going to be merged anytime soon?

PR #311, opened 2019-02-15
Avoid using a loop with the Apt module, because it is deprecated

Rylon:
Fixes the following deprecation warning on Debian-based hosts:

```ansible
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name: {{ item }}`, please use `name: u'{{ mysql_packages 
}}'` and remove the loop. This feature will be removed in version 2.11. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```


geerlingguy:
Thank you; one fewer deprecation warning!

PR #329, opened 2019-06-05
Add `update_password` option to user config

jtickle:
I would like to use Ansible to set up initial user configuration, and manage some configuration after the fact.  But if a user chooses to change their MySQL password, I would like for Ansible to not mess with that.  Ansible 2.0 comes with the `update_password` parameter on `mysql_user` and I find it useful to expose this in the user variables.

Default behavior is left as `always` so existing configurations should not break, but with this you have the option to set it to `on_create` which only sets the password on user creation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ The MySQL users and their privileges. A user has the values:
   - `name`
   - `host` (defaults to `localhost`)
   - `password` (can be plaintext or encrypted—if encrypted, set `encrypted: yes`)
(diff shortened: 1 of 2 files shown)


jtickle:
Not sure why the Travis CI build has been in progress for the length of a human gestation.  Nor why this is stale; in fact I dare say it would be quite useful to many folks who only want Ansible to set their MySQL password on user creation instead of every time the play is run.

divialth:
Is there any chance to reopen this pull request, and getting it merged?
Just had the need for that feature in our deployment.

PR #346, opened 2019-08-13
Fixes #326: Add mysql_character_set_server and mysql_collation_server

### LINKED ISSUE #326: Suggest adding variables for character_set_server and collation_server
These are pretty common options, as utf8mb4 enables emoji support. Additionally, adding the variables to the playbook instead of just using include files to set them enables using them as default when creating database, too.

jeromelefeuvre:


jeromelefeuvre:
Up

PR #351, opened 2019-09-21
Fix mysql tasks when using custom socket

netzvieh:
Update all tasks interacting with the MySQL db on the system to use the
configured socket. Otherwise setting the mysql_socket variable to
something different from the default will result in errors since the
tasks can not connect to the db.

### DIFF
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -2,6 +2,7 @@
 - name: Ensure MySQL databases are present.
   mysql_db:
     name: "{{ item.name }}"
+    login_unix_socket: "{{ mysql_socket }}"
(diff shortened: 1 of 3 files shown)


LordAro:
Can we do better here? There have been at least 4 separate PRs (#312, #351, #358 & #366) that I can see that attempt to fix this issue. Or specifically, ansible/ansible#47736 . Being busy is perfectly fine, but if stalebot is closing things faster than you can triage/merge them, it needs to be reconfigured or removed.

I'm pretty sure this is the cleanest one, and as I understand it the CI should work now that the Debian 10 default changes have been merged.

My specific issue here is the change in behaviour that's brought about by using pymysql as opposed to mysqldb - it no longer defaults to using the socket.

PR #359, opened 2019-10-18
Fix [DEPRECATION WARNING] on Include OS-specific variables (causing A…

senges:
`with_first_found` combined with `skip: true` seems to cause a deprecation warning in `/tasks/variables.ym`l file.
This issue leads to pipeline fail in Azur Devops.

Fixed by fixing `default: ""`


geerlingguy:
This was actually _un_-deprecated — see https://github.com/ansible/ansible/pull/60161

(Background: https://github.com/ansible/ansible/issues/58942).

PR #363, opened 2019-11-19
Add mysql_enable_config_include_dir option

TravisWhitehead:
Previous behavior only includes the `mysql_config_include_dir` in `mysql_config_file` when `mysql_config_include_files` is not empty. This option (when set to true) will always include the `mysql_config_include_dir`.

This is useful for environments where users may have configuration in `/etc/my.cnf.d/` that needs to be enabled, but is not managed by ansible or may be managed elsewhere.

In my opinion `mysql_config_include_dir` should *always* be included regardless of the length of `mysql_config_include_files`, but this PR takes a softer approach in hopes of getting this merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ Whether the global my.cnf should be overwritten each time this role is run. Sett
 
 A list of files that should override the default global my.cnf. Each item in the array requires a "src" parameter which is a path to a file. An optional "force" parameter can force the file to be updated each time ansible runs.
 
+    mysql_enable_config_include_dir: false
+
+Whether `mysql_config_include_dir` should always be included in `mysql_config_file` even when `mysql_config_include_files` is empty.
+
     mysql_databases: []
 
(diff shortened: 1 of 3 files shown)


TravisWhitehead:
@geerlingguy Have you had the opportunity to consider this PR? :)

jeffsheltren:
+1 would love to see this so I don't have to manually work around.

TravisWhitehead:
Given the stale tag, I want to clarify that I'm still hoping this PR can eventually be merged. :)

We have a work around, but it causes `--change` runs in Ansible to issue a false-positive change (because we have to have a separate task to add `!includedir {{ mysql_config_include_dir }}` back to `my.cnf`).

Vancelot11:
+1 this would be super helpful

TravisWhitehead:
This enhancement is still needed.

jeffsheltren:
Bad bot

TravisWhitehead:
@jeffsheltren The bot didn't like that your attitude...

Issue #367, opened 2019-12-05
CentOS 8/RedHat 8 support


mhitza:
While the PR is open I've created a few release on my work fork https://github.com/mhitza/ansible-role-mysql/releases

To use the release I've packaged you can switch out in your Ansible Galaxy file
```
- src: https://github.com/mhitza/ansible-role-mysql/archive/1.2.tar.gz
  name: geerlingguy.mysql
```

I prefer to give it a more obvious name myself, like `ansible-role-mysql-fork-redhat-8` just so it stands out as a workaround.

pulse-mind:
Thank you, please accepte this PR. It is working fine. 
And please publish on galaxy

mhitza:
PR just got merged and it's going to be in the followup release > 2.9.6

geerlingguy:
Fixed, thanks!

PR #368, opened 2019-12-05
Add support for CentOS 8/RedHat 8

mhitza:
Opened up an issue as well #367

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -12,6 +12,7 @@ galaxy_info:
       versions:
         - 6
         - 7
+        - 8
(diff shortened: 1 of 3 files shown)


racke:
I suggest that we add a variable for mysql_python_packages or remove the "Ensure MySQL Python libraries are installed" after adding the Python package to all files in vars.

geerlingguy:
Merged this, but I will be making a few changes to make it so it only installs the Python 3 version for now, as that should be the default for anyone using CentOS 8 (IMO).

PR #371, opened 2019-12-30
Add variable for innodb_buffer_pool_instances

racke:


Joorem:
Definitely needed 👍

simlecp:
I agree with my colleagues, there are a few innodb parameters missing that would be very useful.

innodb_buffer_pool_instances
innodb_flush_method

Issue #332, opened 2019-06-12
Failed to start mariadb service: mysql_pid_file doesn't exist


lifcn:
## My playbook 

```
[xxx@ansible playbooks]$ cat localdev.yaml 
- hosts: dbservers
  become: True
  pre_tasks:
    - name: Install MariaDB Repository
      yum_repository:
        name: mariadb
        description: MariaDB
        baseurl: http://yum.mariadb.org/10.2/centos7-amd64
        gpgkey: https://yum.mariadb.org/RPM-GPG-KEY-MariaDB
        gpgcheck: true
        state: present
      when: ansible_distribution == "CentOS"
    - name: Override variables for MariaDB (RedHat).
      set_fact:
        mysql_packages: ['MariaDB-server', 'MariaDB-client']
        mysql_log_error: /var/log/mysql.err
        mysql_socket: /var/lib/mysql/mysql.sock
      when: ansible_distribution == "CentOS"
  roles:
    - role: geerlingguy.mysql
[xxx@ansible playbooks]$ 
```

> The playbook failed to start mariadb service due to missing mysql_pid_file.   
> Update as below can fix the bug

```
$ diff .ansible/roles/geerlingguy.mysql/tasks/configure.yml .ansible/roles/geerlingguy.mysql/tasks/configure.yml.orig 

54,74d53 
< - name: Create parent directory for mysql_pid_file if it does not exist 
<   file: 
<     path: "{{ mysql_pid_file.split('/')[0:-1]|join('/') }}" 
<     state: directory 
<     recurse: yes 
<     force: no 
<     owner: mysql 
<     group: mysql 
<     mode: 0755 
<     setype: mysqld_db_t 
<  
< - name: Create mysql_pid_file if it does not exist 
<   copy: 
<     content: "" 
<     dest: "{{ mysql_pid_file }}" 
<     force: no 
<     owner: mysql 
<     group: mysql 
<     mode: 0755 
<     setype: mysqld_db_t 
<  
$ 
```

chepec:
Can confirm this exact issue. I ran a playbook with this role on an Ubuntu 18.04 host.

Applying the patch by @lifcn fixed the issue. If I may suggest incorporating this patch into this role @geerlingguy, as it seems to be a simple mistake of omission.

dbeech:
I ran into this issue today. Would be great to see it patched!

PR #382, opened 2020-04-14
Fix issue starting mariadb service due to missing mysql_pid_file

dbeech:
Fixes issue #332 
Patch originally included in a comment by @lifcn, confirmed as working by myself and @chepec.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -51,6 +51,27 @@
     mode: 0755
     setype: mysqld_db_t
 
+- name: Create parent directory for mysql_pid_file if it does not exist
+  file:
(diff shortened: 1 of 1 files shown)


tasks/configure.yml
```diff
@@ -51,6 +51,27 @@
     mode: 0755
     setype: mysqld_db_t
 
+- name: Create parent directory for mysql_pid_file if it does not exist
+  file:
+    path: "{{ mysql_pid_file.split('/')[0:-1]|join('/') }}"
```
Andor, line 56: I don't think you need to create this directory, usually it will be auto-created by init system.
And, for instance, if your pid-file is `/var/lib` you will change ownership for whole `/var/run`

PR #385, opened 2020-04-22
Allow mysql_mode to be set to the empty string

nkakouros:
This is similar to #227 and #229.

According to the [documentation](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_sql_mode), the default for `sql_mode` is **not to be empty**. If someone wants to set it to empty, it is impossible with this role. This PR changes that so that it becomes possible.

I run a custom CMS (not publicly available) that needs that expects that option to be set to empty. 

The PR is minimal and not additional maintenance overhead is added.


nkakouros:
Hope dies last.

nkakouros:
I wonder what it means to run stalebot if there is no engagement by the maintainer. My conclusion is that contributions are not welcome.

nkakouros:
@geerlingguy Could you review this?

nkakouros:
@geerlingguy Could you check this tiny, tiny PR?

nkakouros:
@geerlingguy Are you alive and well?

nkakouros:
😂 🥳 🎉 🎆

geerlingguy:
@nkakouros 😂

Persistence pays off!

PR #391, opened 2020-06-03
Fix replication delegation to Ansible user/host

ncharlot:
Fix replication delegation to Ansible user/host rather than mysql_replication_master that could be an address not reachable when running the playblook.

### DIFF
--- a/tasks/replication.yml
+++ b/tasks/replication.yml
@@ -24,9 +24,14 @@
     - mysql_replication_master | default(false)
   tags: ['skip_ansible_galaxy']
 
+- name: Create group based on replication role.
+  group_by:
+    key: mysql_replication_role_{{ mysql_replication_role }}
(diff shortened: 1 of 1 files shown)


ncharlot:
Hi @geerlingguy! Any chance to have this merge soon?

geerlingguy:
Can you provide some more context, like how this error condition was encountered and how this solves that issue (e.g. steps to replicate)? I would like to try to make sure I understand the problem this fixes and also see if I can set up an environment to replicate that issue.

ncharlot:
The case is when then `mysql_replication_master` is on a private network that local Ansible could not reach.

For example:

    mysql_replication_master: 10.0.0.1  # Lan IP not reachable by local host

The fix aims to find get `ansible_host` but it's not perfect since it will not work if you run a playbook excluding the master.

Issue #344, opened 2019-08-09
Python MySQL dependencies need to be updated


ergonlogic:
On a recent Ubuntu 18.04, we ran into the following fatal error:
```
TASK [geerlingguy.mysql : Remove MySQL test database.] ***************************************************************************
fatal: [test-host]: FAILED! => changed=false 
  msg: The PyMySQL (Python 2.7 and Python 3.X) or MySQL-python (Python 2.X) module is required.
```

ergonlogic:
It looks like this is handled in `tasks/setup-Debian.yml`:

```yaml
- name: Determine required MySQL Python libraries.
  set_fact:
    deb_mysql_python_package: "{% if 'python3' in ansible_python_interpreter|default('') %}python3-mysqldb{% else %}python-mysqldb{% endif %}"

- name: Ensure MySQL Python libraries are installed.
  apt: "name={{ deb_mysql_python_package }} state=present"
```

The [`mysql_user` docs](https://docs.ansible.com/ansible/latest/modules/mysql_user_module.html) state that:

> The below requirements are needed on the host that executes this module.
> 
> PyMySQL (Python 2.7 and Python 3.X), or
> MySQLdb (Python 2.x)

This jives with the error as reported.

I believe that the required package name for the more modern Python (2.7+) should be `python3-pymysql`.

ergonlogic:
Actually, I misread where the error was being generated. It looks like it's the call to `mysql_db` in `tasks/secure-installation.yml`:
```yaml
- name: Remove MySQL test database.
  mysql_db: "name='test' state=absent"
```
The [requirements](https://docs.ansible.com/ansible/latest/modules/mysql_db_module.html#requirements) appear to differ slightly:
```
MySQLdb (Python 2.x)
PyMySQL (Python 2.7 and Python 3.X), or
mysql (command line binary)
mysqldump (command line binary)
```

In our case, we're running Python 2.7, but installing these packages (via `apt`) doesn't appear to resolve the issue. We're starting to suspect that Ansible is using something like `pip freeze` to determine whether these packages are installed.

ergonlogic:
To get this running, we've so far had to upgrade to Python 3. Since Python 2.7 is EOL (on Ubuntu, at least) as of 2020, we're probably just going to have to bite the bullet and move to Python 3.

ergonlogic:
[...]

PR #397, opened 2020-06-26
Fix a fact of discovered interpreter

mtilson:
It seems that "ansible_python_interpreter" is used [to explicitly configure a Python 3 interpreter](https://docs.ansible.com/ansible/latest/reference_appendices/python_3_support.html#using-python-3-on-the-managed-machines-with-commands-and-playbooks).

The gathered fact on the managed host python interpreter could be "discovered_interpreter_python".

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -9,7 +9,7 @@
 
 - name: Determine required MySQL Python libraries.
   set_fact:
-    mysql_python_package_debian: "{% if 'python3' in ansible_python_interpreter|default('') %}python3-mysqldb{% else %}python-mysqldb{% endif %}"
+    mysql_python_package_debian: "{% if 'python3' in discovered_interpreter_python|default('') %}python3-mysqldb{% else %}python-mysqldb{% endif %}"
   when: mysql_python_package_debian is not defined
 
 - name: Ensure MySQL Python libraries are installed.


mtilson:
The issue results in the following on Ubuntu 20.04:
```
TASK [geerlingguy.mysql : Ensure MySQL Python libraries are installed.] *************************************************************************************
fatal: [34.107.110.202]: FAILED! => {"changed": false, "msg": "No package matching 'python-mysqldb' is available"}
```

zeisss:
We are running into the same problem. Working around by manually setting the variable for now.

tisba:
I believe the issue is still around and not addressed, right?

zeisss:
please keep this open

tisba:
- gently ping @geerlingguy -

geerlingguy:
It looks like there's currently a merge conflict. This looks okay to me to merge once that's resolved.

tisba:
Please keep this open, as the issue hasn't been resolved.

tisba:
hmm, I just realised that `tasks/setup-Debian.yml` has changed significantly (see https://github.com/geerlingguy/ansible-role-mysql/blob/a11d7b3cc3df6a528807d66143d9654a556fda0b/tasks/setup-Debian.yml). I understand not enough of ansible and this role to help resolving this :(

zeisss:
@tisba this PR might no longer be necessary given that the python3-mysqldb is now the default: https://github.com/geerlingguy/ansible-role-mysql/issues/344#issuecomment-663119847

geerlingguy:
I'm testing against Ubuntu 20.04 and Rocky Linux 8 now—we'll see how things run on those environments: https://github.com/geerlingguy/ansible-role-mysql/actions/runs/1038672349

geerlingguy:
Yeah; judging by the success of those tests on systems where only Python 3 is present, I think this issue has been resolved in the mean time.

PR #411, opened 2020-10-04
Add mysql_character_set

gparent:
Add mysql_character_set which sets `character-set-server` in the `[mysqld]` section as per: https://dev.mysql.com/doc/refman/5.7/en/charset-applications.html


gparent:
Unstale. Will close if not required once it gets a response.

bmillemathias:
I concur this would be good to have that.

PR #434, opened 2021-02-08
Fix slave check.

iphizic:
Replication user is no set on slave, because use their permission for
check slave status is incorrect.
Fix replication check to root.

### DIFF
--- a/tasks/replication.yml
+++ b/tasks/replication.yml
@@ -15,8 +15,8 @@
 - name: Check slave replication status.
   mysql_replication:
     mode: getslave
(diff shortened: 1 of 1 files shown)


iphizic:
Hello @geerlingguy! Please review this little fix.

iphizic:
@geerlingguy, this fix is nee for mysql master-slave installations.

geerlingguy:
I'm not sure why this would be the case—if you have a mysql_replication_user set with the proper permissions, then that should work...

Issue #435, opened 2021-02-10
Getslave operation always fails while configuring the replication slave


blajos:
Mysql_replication_user is not created on the replication slave, so this task fails reliably:
https://github.com/geerlingguy/ansible-role-mysql/blob/3a164a5d0867e7b5e85e067fb260d21b1132213e/tasks/replication.yml#L15

I think the solution is to remove login_user and login_password from the task.

dorianm:
Same problem here. I try with your fix (#436) and it works ! Thank you !

PR #436, opened 2021-02-10
Remove login_user and login_password from getslave operation

blajos:
Replication user is not created on the slave, but root works just fine.

Should fix https://github.com/geerlingguy/ansible-role-mysql/issues/435


blajos:
This PR is still valid.

PR #437, opened 2021-02-20
Document that the non-root user password can be updated

pfrenssen:
I am using a non-root user to manage the database and I wanted to change this user's password but it was not immediately clear from the documentation how to do this.


pfrenssen:
Still valid. Failures are not caused by this PR.

pfrenssen:
Not stale.

PR #446, opened 2021-04-23
Issue #438: Install python3-pymysql on Debian 10.

### LINKED ISSUE #438: Module fails to interact with database due to missing python 2 module
Python interpreter: Python 2

Affected OS:
- Debian 9
- Debian 10
- Ubuntu 18.04

I took the role as is and ran it against the server which got me the following message:

    TASK [geerlingguy.mysql : Remove MySQL test database.] **************************************************************
    fatal: [server]: FAILED! => {"changed": false, "msg": "The PyMySQL (Python 2.7 and Python 3.X) or MySQL-python (Python 2.X) module is required."}

As far as I understand the role should make sure the proper Python modules are installed, right?
I build my own role some time ago and took care of that but I can not figure out, why this does not work here.
Any hint to what I am doing wrong?

geerlingguy:
Fixes #438.

### DIFF
--- a/vars/Debian-10.yml
+++ b/vars/Debian-10.yml
@@ -2,6 +2,7 @@
 __mysql_daemon: mariadb
 __mysql_packages:
   - default-mysql-server
+  - python3-pymysql
 mysql_log_file_group: adm
 __mysql_slow_query_log_file: /var/log/mysql/mysql-slow.log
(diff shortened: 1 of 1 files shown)


jwarnier:
I believe this might be fixed by https://github.com/geerlingguy/ansible-role-mysql/commit/48121b41b561bd8fb21fe5cd2c19dce0a6857ab1

PR #449, opened 2021-05-26
Add TLS settings to mysql.cnf

tmgstevens:
Add additional settings for TLS/SSL configuration into mysql.cnf (reference: https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.htm)

### DIFF
--- a/templates/my.cnf.j2
+++ b/templates/my.cnf.j2
@@ -18,6 +18,35 @@ skip-name-resolve
 sql_mode = {{ mysql_sql_mode }}
 {% endif %}
 
+# SSL configuration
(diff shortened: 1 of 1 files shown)


hectoralicea:
this looks like a very good idea especially in agencies that are getting dinged for having software that support TLS 1.1 encryption

tmgstevens:
Can we get a commit on this please?

tmgstevens:
Please could you commit this @geerlingguy ?

tmgstevens:
This is still relevant - can we get a merge please?

PR #457, opened 2021-08-17
Debian 11 support

### LINKED ISSUE #469: missing Debian 11 vars
like defined in _tasks/variables.yml_ in line 7:
```
---
# Variable configuration.
- name: Include OS-specific variables.
  include_vars: "{{ item }}"
  with_first_found:
    - files:
        - "vars/{{ ansible_os_family }}-{{ ansible_distribution_major_version }}.yml"
        - "vars/{{ ansible_os_family }}.yml"
      skip: true
```
the variables will be used for the major distibution version first, but in the vars directory there is only _Debian.yml_ and _Debian-10.yml_.
The definitions for Debian-11 are missing. So the task "Ensure MySQL packages are installed" is failing with:
`FAILED! => {"changed": false, "msg": "No package matching 'mysql-server' is available"}`

My workaround for this is at the moment to create a symlink to the _Debian-10.yml_ with:
`ln -s Debian-10.yml Debian-11.yml`

smallsam:
Debian 11/bullseye vars and adding to ci

Fixes #469

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
           - ubuntu2004
           - ubuntu1804
           - debian10
+          - debian11
 
     steps:
       - name: Check out the codebase.
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -1,6 +1,6 @@
 ---
 - name: Check if MySQL is already installed.
-  stat: path=/etc/init.d/mysql
+  stat: path=/etc/init.d/{{ mysql_daemon }}
   register: mysql_installed
 
 - name: Update apt cache if MySQL is not yet installed.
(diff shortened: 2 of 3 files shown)


DO1JLR (approved):

SchizoDuckie:
As far as I can tell, the check in `setup-Debian.yml` will fail as the default package installs mariadb, which does NOT create a file in /etc/init.d, leading to idempotency errors and deleted innodb logfiles because the role thinks it's always just installed the service.

suggested fix: 

```YAML 
#setup-Debian.yml
- name: Check if MySQL is already installed.
  stat: "path=/etc/init.d/{{ mysql_daemon }}"
  register: mysql_installed
```

```YAML 
#Debian11.yaml
__mysql_daemon: mariadb
```

domrim:
can confirm @SchizoDuckie . without these changes the role doesn't run idempotent, with them it does. can sb fix this and get this ready to merge? Debian 11 gets more important every day for us.

smallsam:
Thanks @SchizoDuckie and @domrim I've incorporated this fix here now.

Issue #223, opened 2017-06-13
Default insecure passwords seems bad


eythian:
I think that setting a default password of 'root' is dangerous. Instead, it might be worth considering:

* aborting with an error if no password is defined
* setting a large random one if none is defined

That should reduce the risk of people ending up with an insecure setup because they were in a hurry or something. Relatedly, it could be interesting to have a little aside in the docs on how to securely create a password file for when you drop this role in as a git submodule, and so want to avoid making changes to it if possible.

geerlingguy:
> I think that setting a default password of 'root' is dangerous.

At least it's more secure than MySQL's traditional default of having no password :)

I've considered doing something like this from time to time... but it's difficult to think of a way to randomize it (where it's random on first run, then consistent on future runs), and it's also just security-through-obscurity to simply set a more secure default.

Throwing a warning if it's `root`, I might be amenable to, since it would be the least invasive and most reliable way of highlighting insecure passwords.

But one thing to keep in mind is that many people (myself included) end up using the defaults for things like one-off database environments where we need to build something quick, run some commands, then tear it down. I usually either use `'root'` or `''` (no password) for that, just because it's fast and works quickly with defaults in many of the downstream tools I use with MySQL...

eythian:
Presumably the root password is only set if the .my.cnf exists (excluding that override option), so if a random one were generated, it would be set and written into that file the first time, and wouldn't change again?

But even a loud warning would be enough I think.

Thanks for the role btw, it's a very handy thing for me finally getting my wordpress installation ansibleised :)

llbbl:
maybe optionally prompting for a password based on config parameter. this would not break it for people just needing an easy way to set a password and still allow for a bit more control over access to password.

colans:
@geerlingguy wrote:
> I've considered doing something like this from time to time... but it's difficult to think of a way to randomize it (where it's random on first run, then consistent on future runs), and it's also just security-through-obscurity to simply set a more secure default.

Agreed on those points.  The solution here is to do the following:

1. Set a random password on each run (unless there's already one there and there's no reason to change it).
1. Place the generated password in `/root/.my.cnf`.

> But one thing to keep in mind is that many people (myself included) end up using the defaults for things like one-off database environments where we need to build something quick, run some commands, then tear it down. I usually either use 'root' or '' (no password) for that, just because it's fast and works quickly with defaults in many of the downstream tools I use with MySQL...

[...]

PR #465, opened 2021-10-07
Issue #223: Generate a random default root password

colans:
Instead of hard-coding an insecure default password for the `root` MySQL user, generate a random one as discussed in issue #223.

This is more complicated than simply adding a random password to the defaults file because that'll change every time the variable is read.  Instead, I generate a random password and place it into `/root/.my.cnf` (or wherever) unless there already is one, fetch it from there, and *then* use it to set the actual `root` password in the DB.  But this only happens if the configuration file has changed.  Otherwise, they'd be out of sync.


colans:
@geerlingguy I believe that this is ready to be merged.  Or do you have any additional concerns?

colans:
:wave:

colans:
:wave:

geerlingguy:
Stale bot stay away!

PR #474, opened 2022-01-12
Disable Starting of MySQL on Debian during install.

misilot:
Disable Starting of MySQL Server on Debian based systems during install.

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -17,6 +17,7 @@
   apt:
     name: "{{ mysql_packages }}"
     state: present
+    policy_rc_d: 101
   register: deb_mysql_install_packages
 
 # Because Ubuntu starts MySQL as part of the install process, we need to stop


misilot:
not stale

misilot:
not stale

geerlingguy:
I like this idea, but for some reason tests aren't running on this PR...?

misilot:
Let me rebase and see what happens

JoyceBabu:
I am trying to install MariaDB 10.8 with this role by adding the mariadb repository for the version using a pre task. But the role is failing when running task [Ensure MySQL is started and enabled on boot](https://github.com/geerlingguy/ansible-role-mysql/blob/master/tasks/configure.yml#L85) since the default `ib_logfile0` was deleted by the role after installation.

> [ERROR] InnoDB: File ./ib_logfile0 was not found

Will this PR fix the issue?

johansenha:
> I am trying to install MariaDB 10.8 with this role by adding the mariadb repository for the version using a pre task. But the role is failing when running task [Ensure MySQL is started and enabled on boot](https://github.com/geerlingguy/ansible-role-mysql/blob/master/tasks/configure.yml#L85) since the default `ib_logfile0` was deleted by the role after installation.
> 
> > [ERROR] InnoDB: File ./ib_logfile0 was not found
> 
> Will this PR fix the issue?

same happening to me. Mariadb 10.8 on ubuntu22.04.01

misilot:
Thanks @geerlingguy!

PR #476, opened 2022-01-21
This makes it work with Rocky.

itd:


itd:
yup. tested. works. Weird.

geerlingguy:
Can you also add in `rockylinux8` to the .github/workflows/CI.yml list of tested OSes?

kbendl:
Looks like rockylinux8 was already in the ci.yml, but I fixed trailing whitespace in the Rocky8.yml file that was causing the ci to fail. Thanks!

ealekseev:
Hi! What should happen to have this merged? Looks like it failed on tests not related to change itself.

tasks/Rocky.yml
```diff
@@ -0,0 +1,7 @@
+---
```
geerlingguy, line 1: This file would be better named `setup-Rocky.yml`, but I'd rather make it so the same `setup-RedHat` is run for Rocky, since it's the same base... https://github.com/geerlingguy/ansible-role-mysql/blob/master/tasks/setup-RedHat.yml

Additionally, I would like to make sure rockylinux8 is added to `.github/workflows/CI.yml`.

geerlingguy (changes requested):

tasks/Rocky.yml
ealekseev: Hi Jeff! Thx!
Sorry, I didn't work with github a lot recently. It seems I can't contribute to someone else's PR?
What is the best approach here - me to create my own fork and PR?

PR #478, opened 2022-01-26
Fix idempotence for Debian 11 and Debian 10 tests

geerlingguy:
It seems that #457 may have broken idempotence on Debian 10. Testing to see if that's the case. Don't want to tag broken code in a new release...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,11 @@ jobs:
       matrix:
         distro:
           - rockylinux8
+          - centos7
+          - ubuntu2204
           - ubuntu2004
           - debian11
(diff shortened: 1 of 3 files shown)


geerlingguy:
Heh... so idempotence breaks on debian 10 without the change, but breaks on debian 11 with the change. Drat.

geerlingguy:
What is that check even for? Maybe a better check is needed. It's been a while, but I think it's checking for an initv script... which is probably not even a thing on Debian 11.

ak4t0sh:
FYI just checked on a fresh install it seems that the file has been renamed to /etc/init.d/mariadb on debian11

PR #494, opened 2022-07-27
Update replication.yml

Aguay-val:
Mode name has changed. Here the patch to make this work again.

### DIFF
--- a/tasks/replication.yml
+++ b/tasks/replication.yml
@@ -14,7 +14,7 @@
 
 - name: Check slave replication status.
   mysql_replication:
-    mode: getslave
+    mode: getreplica
     login_user: "{{ mysql_replication_user.name }}"
     login_password: "{{ mysql_replication_user.password }}"
   ignore_errors: true
@@ -25,7 +25,7 @@
   tags: ['skip_ansible_galaxy']
 
 - name: Check master replication status.
-  mysql_replication: mode=getmaster
+  mysql_replication: mode=getprimary
   delegate_to: "{{ mysql_replication_master }}"
   register: master
   when:
@@ -36,7 +36,7 @@
 
 - name: Configure replication on the slave.
   mysql_replication:
-    mode: changemaster
+    mode: changeprimary
     master_host: "{{ mysql_replication_master }}"
     master_user: "{{ mysql_replication_user.name }}"
     master_password: "{{ mysql_replication_user.password }}"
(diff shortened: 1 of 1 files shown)


Aguay-val:
Since 3.0.0 https://github.com/ansible-collections/community.mysql/blob/main/CHANGELOG.rst the "master" and "slave" support has been droped. So now it's replica and primary.

geerlingguy:
@Aguay-val - Do you know which version of Ansible (community edition, like 4.x, 5.x, or 6.x) the 3.0.0 version was included in? I don't want to merge this in until I'm sure it won't affect existing users on slightly older supported versions of Ansible.

It looks like 3.0.0. was only released in December (https://github.com/ansible-collections/community.mysql/releases/tag/3.0.0), so it is likely many people are running older versions of the collection if they're not managing their dependencies themselves.

I do want to merge this to make sure the role stays up to date though. I will have to bump the major version since it's a breaking change.

Aguay-val:
Since 2.0.0 the use of "replica" and "primary" has been introduced : https://github.com/ansible-collections/community.mysql/blob/2.0.0/changelogs/CHANGELOG.rst#id23
It has been working until 3.0.0 because the python script has dropped the support of "master" and "slave" after the 3.0.0 release : https://github.com/ansible-collections/community.mysql/commit/1f796d9e812f77ec988234e120f88695bf65fd9a

I don't know when the 3.0.0 has been included in ansible. I'll check this out and come back later with more informations.

Aguay-val:
> Since 2.0.0 the use of "replica" and "primary" has been introduced : https://github.com/ansible-collections/community.mysql/blob/2.0.0/changelogs/CHANGELOG.rst#id23 It has been working until 3.0.0 because the python script has dropped the support of "master" and "slave" after the 3.0.0 release : [ansible-collections/community.mysql@1f796d9](https://github.com/ansible-collections/community.mysql/commit/1f796d9e812f77ec988234e120f88695bf65fd9a)
> 
> I don't know when the 3.0.0 has been included in ansible. I'll check this out and come back later with more informations.

If I'm misunderstanding, since ansible 4.x `community.mysql` is using 2.1.0 which was already using an alternative to "master/slave" with "Primary/replica" https://github.com/ansible-community/ansible-build-data/blob/main/4/ansible-4.0.0.deps#L39

PR #495, opened 2022-08-04
Use root user to check replica status

jlestavel:
If I'm not wrong, to check the replica status, we can't use the `mysql_replication_user` user, as it's only created on the primary MySQL instance (and here we're on the replica MySQL instance, with the `mysql_replication_role == 'slave'` condition). Shouldn't we use the `root` user? Or did I miss something?


geerlingguy:
The replication user would be replicated across the instances, correct? Are you thinking of the root user in linux, or in the MySQL db?

jlestavel:
> The replication user would be replicated across the instances, correct?

For me, the replication user is only created on the primary DB, cf

https://github.com/geerlingguy/ansible-role-mysql/blob/8ae4de3e9f518bd855bebd33aad1b2ca4c7e3b1c/tasks/replication.yml#L2-L10

> Are you thinking of the root user in linux, or in the MySQL db?

I'm thinking of the MySQL `root` user

zangra-dev:
This seems to fix and correctly execute the task:
- name: Check slave replication status.

jlestavel:
@geerlingguy I fixed the conflict with `master`. Could you have a look please?

PR #498, opened 2022-08-12
Make replication read_only configurable

clementd64:
Make replication read_only optionnal


templates/my.cnf.j2
```diff
@@ -63,7 +63,9 @@ binlog_ignore_db = {{ db.name }}
 {% endif %}
 
 {% if mysql_replication_role == 'slave' %}
+{% if mysql_replication_read_only | default(true) %}
```
geerlingguy, line 66: Can you also add the default variable to the default vars file, and document it in the README with the other replication-related vars? That way you could also pull off the `| default(true)` from here since it will be defined in role defaults.

PR #503, opened 2022-11-17
Allow replica master to be differ from local inventory + fix Is_Replica or Is_Slave

giorgioma:
E.g.
I am running ansible on my local machine and they are accessible over public IP but i want the master and the slave to communicate over some other network interface card.

`mysql_replication_master_inventory_host` defaults to `mysql_replication_master` so it's a non-breaking change

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Replication settings. Set `mysql_server_id` and `mysql_replication_role` by serv
 
 `mysql_replication_master` needs to resolve to an IP or a hostname which is accessable to the Slaves (this could be a `/etc/hosts` injection or some other means), otherwise the slaves cannot communicate to the master.
 
-If the replication master has different IP addresses where you are running ansible and where the mysql replica is running, you can *optionally* specify a `mysql_replication_master_inventory_host` to access the machine (e.g. you run ansible on your local machine, but the mysql master and replica need to communicate behind a vpn)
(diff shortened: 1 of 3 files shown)


giorgioma:
I have upgraded some dependencies in my dev environment (MariaDB) and my slave data from getreplica is now
```
ok: [db2] => {
    "slave": {
        "Is_Replica": false,
        "changed": false,
        "failed": false,
        "msg": "Server is not configured as mysql replica",
        "queries": []
    }
}
```
but the current check is 
`(slave.Is_Slave is defined and not slave.Is_Slave) or (slave.Is_Slave is not defined and slave is failed)`
I am guessing it should now be
`(slave.Is_Slave is defined and not slave.Is_Slave) or (slave.Is_Replica is defined and not slave.Is_Replica) or (slave.Is_Slave is not defined and slave.Is_Replica is not defined and slave is failed)`
to cover both cases for the naming transition

liammcdermott:
Hey @giorgioma, I'm just wonder if there's a reason the task: `- name: Start replication.` didn't get the same treatment? It still has the old behaviour of only checking `slave.Is_Slave`.

I notice the logic of that `when` might be incorrect, is that the reason you didn't touch it?

hahayusuf:
I am having the same issue of Start replication never triggering. Not sure why this is the case.

liammcdermott:
@hahayusuf I _think_ what's wrong with it is mentioned in my edit. `not save.Is_Replica` evaluates to `false`, since the replication is setup but not started. I tested and at least this was the case for me.

giorgioma:
Ah yes! I was informed i broke my staging env as well, i'll open a pr in the morning
I simply didn't notice that the last command had that condition as well

liammcdermott:
@giorgioma No worries, I created #511. Would you mind trying that out and seeing if it works for you?

PR #507, opened 2022-11-24
Remove `warn` as it disappeared in Ansible 7

eRadical:
More at https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_7.html#removed-features


geerlingguy:
Indeed. Thanks!

Issue #415, opened 2020-10-23
mysql 8 comes with mysqlx plugin listening on all interfaces by default


zeisss:
Not sure if this something you want to fix in the your defaults, but this was quite surprising to us:

Mysql 8 comes with a plugin called "mysql x" that openes a second server socket (port * 10 by default) on all interfaces (in our case: the public internet).

Given that this role sets the `bind-address` by default to localhost, it might also wants to set the `mysqlx-bind-address` to something more restrictive and/or disable it default.

Docs: 
* https://dev.mysql.com/doc/refman/8.0/en/x-plugin-options-system-variables.html
* https://dev.mysql.com/doc/refman/8.0/en/x-plugin-disabling.html

fama:
I second that.  I wish there was a way to disable the mysqlx with ansible variable to add to config file
`mysqlx=0` as this is certainly a security risk:

```
ss -tulpn
Netid     State      Recv-Q     Send-Q         Local Address:Port           Peer Address:Port     Process     
...
tcp       LISTEN     0          151                127.0.0.1:3306                0.0.0.0:*                    
tcp       LISTEN     0          70                         *:33060                     *:*                    
...
```

nathan-sav:
This should definitely get fixed, please reopen! @geerlingguy

sebcmp:
Ubuntu/Debian defaults to localhost

```bash
> grep bind /etc/mysql/mysql.conf.d/mysqld.cnf 
bind-address		= 127.0.0.1
mysqlx-bind-address	= 127.0.0.1
```

This role should really default to 127.0.0.1 on both bind addresses instead of 0.0.0.0.

dw-ec:
I managed to work around this using this role's built-in config override mechanism by setting a variable:

```
 mysql_config_include_files:
   - src: mysql-extra-cnf/mysqlx.cnf
```

Then creating a template like `templates/mysql-extra-cnf/mysqlx.cnf`:

```
[mysqld]
mysqlx-bind-address={{ mysqlx_bind_address | default( mysql_bind_address ) }}
```

PR #509, opened 2022-12-03
Add mysqlx variables to my.cnf

bratucornel:
This change let you customize the `mysqlx-port` and `mysqlx-bind-address` in order to stop mysqlx plugin to listen on all interfaces. Fix for #415


sebcmp:
Might wanna add `mysqlx = OFF` option to turn it off altogether.

PR #511, opened 2022-12-13
Fix replication not starting.

liammcdermott:
Fofllow-up to #503. That PR didn't include `Is_Replica` in its `Start replication.` task.

In addition I found that the previous logic of running `Start replication.` when `not slave.Is_Slave` would never start the replica, since `slave.Is_Slave` would evaluate to `true`. The reason being the replication _is_ setup, it just hasn't been started yet, so this PR also fixes that.


giorgioma:
Works for me

PR #514, opened 2023-02-11
Fix no-free-form rule violation

ssbarnea:
Related: https://ansible-lint.readthedocs.io/rules/no-free-form/


geerlingguy:
LGTM. I've slowly been adjusting my shorthand use over the years, but as this role is probably my oldest (or nearly so), it still had some lurking there. Thanks!

PR #518, opened 2023-02-23
Initialize Custom Data Directory

### LINKED ISSUE #510: MySQL fails to start when `mysql_datadir` is set to anything other than MySQL's default
Our Ansible playbooks use this role and set the MySQL datadir like this:

```
mysql_datadir: /data/mysql
```

When this is set the plays fail with this error on every database server:

```
fatal: [db05]: FAILED! => {"changed": false, "msg": "Unable to start service mysql: Job for mysql.service failed because the control process exited with error code.\nSee \"systemctl status mysql.service\" and \"journalctl -xeu mysql.service\" for details.\n"}
```

Then running `sudo journalctl -xeu mysql` turns up the following:

```
mysql-systemd-start[450456]: MySQL system database not found in /data/mysql. Please run mysqld --initialize
```

Unfortunately, running `mysqld --initialize` appears to work, but causes plays to fail at a later step with a MySQL permissions error. So it's not a viable workaround.

This issue is a duplicate of #399 , the bot closed it, despite it being a current issue.

OS: Ubuntu 22.04
MySQL version: `/usr/sbin/mysqld  Ver 8.0.31-0ubuntu0.22.04.1 for Linux on x86_64 ((Ubuntu))`

misilot:
Partially Fixes #510

This will create the expected data directory. This most likely does not address protections in place because of AppArmor on Ubuntu

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -80,6 +80,18 @@
     - mysql_log_error | default(false)
   tags: ['skip_ansible_galaxy']
 
+- name: Initialize mysql when mysql_datadir is customized (MySQL)
+  ansible.builtin.command: mysqld --initialize-insecure --user=mysql --datadir={{ mysql_datadir }}
+  args:
+    creates: "{{ mysql_datadir }}/mysql"
+  when: mysql_datadir != "/var/lib/mysql" and ('5.7.' in mysql_cli_version.stdout or '8.0.' in mysql_cli_version.stdout)
+
+- name: Initialize mysql when mysql_datadir is customized (MariaDB)
+  ansible.builtin.command: mysql_install_db
+  args:
+    creates: "{{ mysql_datadir }}/mysql"
+  when: mysql_datadir != "/var/lib/mysql" and ('5.7.' not in mysql_cli_version.stdout or '8.0.' not in mysql_cli_version.stdout)
+
 - name: Ensure MySQL is started and enabled on boot.
   ansible.builtin.service:
     name: "{{ mysql_daemon }}"
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -3,6 +3,9 @@
   hosts: all
   become: true
 
+  vars:
+    mysql_datadir: "/opt/mysql"
+
   roles:
     - role: geerlingguy.mysql
 


geerlingguy:
Since I don't customize the dir, I'd like someone else to confirm this works / test it out (preferably across a couple platforms) before merging. Also, it might be nice to test it in CI.

misilot:
Any tips or examples of how we might add a test to CI?

Thanks!

widhalmt:
Hi. I just ran into the same issue. I *do* have to change the data dir. So I'll check the change. And I can help with the CI part, too.

widhalmt:
Was easier than expected since there's only one molecule scenario. @geerlingguy I'll post a PR shortly with setting the variable in this one scenario. Is that what you thought of? Or would you like to have an extra scenario to test changed variables? I saw your default scenario is quite straight forward so maybe you want to keep it that way and have another one for more peculiar setups?

widhalmt:
@misilot I tried your solution on Ubuntu 20.04 with MariaDB 10.3.38. It created the usual InnodDB file but not the default database like the `mysql` one. It only worked when I ran `mysqld_install_db`. Maybe you need to change your Merge Request. If it worked for you, I need to dig deeper.

misilot:
@widhalmt let me take a look and try again.

misilot:
@widhalmt so MariaDB and MySQL have different ways to initialize the data directory. I added a conditional to check if mariadb-server exists in the `mysql_packages` variable.

tasks/configure.yml
```diff
@@ -80,6 +80,18 @@
     - mysql_log_error | default(false)
   tags: ['skip_ansible_galaxy']
 
+- name: Initialize mysql when mysql_datadir is customized (MySQL)
+  ansible.builtin.command: mysqld --initialize-insecure --user=mysql --datadir={{ mysql_datadir }}
```
geerlingguy, line 84: This could interfere with later initialization, though, no?

misilot: Which later initialization? I can't get molecule to run the rockylinux on my m1 mac, but the ubuntu2204 and 2004, both seem to pass the tests, when specifying an alternate data directory for both mysql and mariadb. 


I am trying to modify the github actions to run the additional tests (I believe that is the way to do it?) I can't find a way to have different playbooks in molecule that tear down everything between playbooks.

widhalmt: It still depends. @geerlingguy do you want to have an extra molecule scenario for a moved directory or should it be part of `default`? I made a PR to the PR of @misilot and I'm ready to change as you want it to have.

misilot: So, I am working on the logic to check the version number instead of looking for package names, since debian uses mysql in the package name, but is really mariadb, the logic I have doesn't work

misilot: Ok, I fixed the logic, and they all seem to run locally with molecule. I also added in @widhalmt change to this PR.

geerlingguy:
This is currently failing, but might just need a rebase?

misilot:
I just rebased, so we will see what it does.

PR #528, opened 2023-07-19
Replace ansible.builtin.command with command module

sanjeetsagar:
The `ansible.builtin.command` module was removed in Ansible 2.9.6. This PR replaces the use of `ansible.builtin.command` with the `command` module

The `command` module is a more generic module that can be used to execute any command on a remote machine. It has a number of additional features that were not available in the `ansible.builtin.command` module.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,6 +1,6 @@
 ---
 - name: restart mysql
-  ansible.builtin.service:
+  service:
     name: "{{ mysql_daemon }}"
     state: restarted
     sleep: 5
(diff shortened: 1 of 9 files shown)


SuperQ:
Your Ansible is broken, `ansible.builtin.command` is very much a thing.

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html

geerlingguy:
Yes; you just need to make sure you're running a newer version of Ansible and this will work correctly.

PR #532, opened 2023-11-28
Hide passwords during execution

ptomas-adacis:
Hi,

This small pull request allows user to hide current plain-text passwords into execution output with the new variable `mysql_hide_passwords`.
The default value of this variable is `false` to allow users to activate this feature if they need it with a minimal impact on their current usage.

Thank you.

PR #533, opened 2023-12-06
Update replication.yml to use "| default(omit, true)" for delegate_to

pgehres:
Looks like this was a change in ansible-core 2.16 which results in an error like 

`
fatal: [10.63.4.102 -> {{ mysql_replication_master }}]: FAILED! => {"msg": "Supplied entity must be Host or Group, got <class 'ansible.inventory.host.Host'> instead"}
`

Adding the suggested fix per https://github.com/ansible/ansible/issues/82264

### DIFF
--- a/tasks/replication.yml
+++ b/tasks/replication.yml
@@ -27,7 +27,7 @@
 - name: Check master replication status.
   mysql_replication:
     mode: getprimary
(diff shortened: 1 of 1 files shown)


geerlingguy:
This has been fixed in https://github.com/geerlingguy/ansible-role-mysql/pull/532 - thanks!

PR #541, opened 2024-05-10
Set the default Error Log location for Ubuntu

### LINKED ISSUE #540: Update logrotate in Ubuntu for mysql.err
The error messages in at least Ubuntu 20.04 for mysql get logged to syslog, so the mysql.err log is not handled by logrotate when this role is installed.

looking at /etc/logrotate.d/mysql-server is mentions 
> The error log is obsolete, messages go to syslog now.

```
# - I put everything in one block and added sharedscripts, so that mysql gets
#   flush-logs'd only once.
#   Else the binary logs would automatically increase by n times every day.
# - The error log is obsolete, messages go to syslog now.
/var/log/mysql.log /var/log/mysql/*log {
	daily
	rotate 7
	missingok
	create 640 mysql adm
	compress
	sharedscripts
	postrotate
		test -x /usr/bin/mysqladmin || exit 0
		# If this fails, check debian.conf!
		MYADMIN="/usr/bin/mysqladmin --defaults-file=/etc/mysql/debian.cnf"
		if [ -z "`$MYADMIN ping 2>/dev/null`" ]; then
		  # Really no mysqld or rather a missing debian-sys-maint user?
		  # If this occurs and is not a error please report a bug.
		  #if ps cax | grep -q mysqld; then
		  if killall -q -s0 -umysql mysqld; then
 		    exit 1
		  fi
		else
		  $MYADMIN flush-logs
		fi
	endscript
}
```

misilot:
This allows it to be managed by the default logrotate configuration

Fixes #540

### DIFF
--- a/vars/Ubuntu.yml
+++ b/vars/Ubuntu.yml
@@ -1,0 +1,2 @@
+---
+__mysql_log_error: /var/log/mysql/error.log


misilot:
no stale

PR #549, opened 2024-08-18
Support MySQL 8.4

exileed:
I've changed the method for resetting the root user password in MySQL 8+ since `mysql_native_password` is no longer enabled by default. Link https://dev.mysql.com/doc/refman/8.4/en/native-pluggable-authentication.html

Now, versions of MySQL 8.* are correctly identified.

### DIFF
--- a/tasks/secure-installation.yml
+++ b/tasks/secure-installation.yml
@@ -50,7 +50,7 @@
   with_items: "{{ mysql_root_hosts.stdout_lines|default([]) }}"
   when: >
     ((mysql_install_packages | bool) or mysql_root_password_update)
-    and ('5.7.' in mysql_cli_version.stdout or '8.0.' in mysql_cli_version.stdout)
+    and ('5.7.' in mysql_cli_version.stdout or '8.' in mysql_cli_version.stdout)
 
 # Set root password for MySQL < 5.7.x.
 - name: Update MySQL root password for localhost root account (< 5.7.x).
@@ -61,7 +61,7 @@
   with_items: "{{ mysql_root_hosts.stdout_lines|default([]) }}"
   when: >
     ((mysql_install_packages | bool) or mysql_root_password_update)
-    and ('5.7.' not in mysql_cli_version.stdout and '8.0.' not in mysql_cli_version.stdout)
+    and ('5.7.' not in mysql_cli_version.stdout and '8.' not in mysql_cli_version.stdout)
 
 # Has to be after the root password assignment, for idempotency.
 - name: Copy .my.cnf file with root password credentials.
--- a/tasks/secure-installation.yml
+++ b/tasks/secure-installation.yml
@@ -41,6 +41,18 @@
 # Note: We do not use mysql_user for this operation, as it doesn't always update
 # the root password correctly. See: https://goo.gl/MSOejW
 # Set root password for MySQL >= 5.7.x.
+
+- name: Update MySQL root password for localhost root account (8.x).
+  ansible.builtin.shell: >
+    mysql -u root -NBe
+    "ALTER USER '{{ mysql_root_username }}'@'{{ item }}'
+     IDENTIFIED WITH caching_sha2_password BY '{{ mysql_root_password }}'; FLUSH PRIVILEGES;"
+  no_log: "{{ mysql_hide_passwords }}"
+  with_items: "{{ mysql_root_hosts.stdout_lines|default([]) }}"
+  when: >
+    ((mysql_install_packages | bool) or mysql_root_password_update)
+    and ('8.' in mysql_cli_version.stdout)
+
 - name: Update MySQL root password for localhost root account (5.7.x).
   ansible.builtin.shell: >
     mysql -u root -NBe
@@ -50,7 +62,7 @@
   with_items: "{{ mysql_root_hosts.stdout_lines|default([]) }}"
   when: >
     ((mysql_install_packages | bool) or mysql_root_password_update)
-    and ('5.7.' in mysql_cli_version.stdout or '8.' in mysql_cli_version.stdout)
+    and ('5.7.' in mysql_cli_version.stdout)
 
 # Set root password for MySQL < 5.7.x.
 - name: Update MySQL root password for localhost root account (< 5.7.x).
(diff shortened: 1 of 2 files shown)


tasks/secure-installation.yml
```diff
@@ -61,7 +73,7 @@
   with_items: "{{ mysql_root_hosts.stdout_lines|default([]) }}"
   when: >
     ((mysql_install_packages | bool) or mysql_root_password_update)
-    and ('5.7.' not in mysql_cli_version.stdout and '8.0.' not in mysql_cli_version.stdout)
+    and ('5.7.' not in mysql_cli_version.stdout and '8.' not in mysql_cli_version.stdout)
```
geerlingguy, line 76: The only downside to this change is maybe the comparison would break if there were a version like `5.8.0`, that version would be identified as `8.`... trying to think of a better way to narrow it. Maybe add in something to trim off the first two digits, and compare that to `8.`?

exileed: I was thinking about this solution, but as far as I know, version 5.8 doesn't exist. However, everything works fine on version >= 5.7. If you want, I can change the version definition using trim.

geerlingguy: I think using trim would be ever so slightly more future proof. It feels more correct to me.

tmace85:
Hi Guys, 

thanks for your awsome work. i would appreciate if this got released in near future :)  I´ve successully tested with mysql 8.4 and ubuntu 24.04.

Let me know if i can assist. 

Cheers Tim

tasks/secure-installation.yml
Diegiwg, line 76:
```suggestion
  and (mysql_cli_version.stdout is version('5.7', '<'))
```

I suggest using the [ansible.builtin.version_test](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html) filter to validate version conditions within each task.

This ensures a more standardized and reliable approach to version checking.

@exileed @geerlingguy

Diegiwg: https://github.com/geerlingguy/ansible-role-mysql/pull/549#discussion_r1957163115

Diegiwg: Simple Molecule Test to Demonstrate the Functionality of the Suggested Module:


```yaml
---
- name: Validate MySQL Version conditions
  hosts: all
  tasks:
    - name: "Set mock MySQL version to 5.6.x (pre-5.7 boundary test)"
      ansible.builtin.set_fact:
        mysql_version: "5.6.51-extra-ctx"

    - name: "Validate 5.6.x version is recognized as pre-5.7"
      ansible.builtin.assert:
        that: "{{ mysql_version is version('5.7', '<') }}"

    - name: "Set mock MySQL version to 8.4.x (post-8.0 boundary test)"
      ansible.builtin.set_fact:
        mysql_version: "8.4.4-extra-ctx"

    - name: "Validate 8.x version is recognized as post-8.0"
      ansible.builtin.assert:
        that: "{{ mysql_version is version('8.0', '>=') }}"

```

image

templates/my.cnf.j2
```diff
@@ -104,7 +104,7 @@ lower_case_table_names = {{ mysql_lower_case_table_names }}
 event_scheduler = {{ mysql_event_scheduler_state }}
 
 # InnoDB settings.
-{% if mysql_supports_innodb_large_prefix and '8.0.' not in mysql_cli_version.stdout %}
+{% if mysql_supports_innodb_large_prefix and '8.' not in mysql_cli_version.stdout %}
```
Diegiwg, line 107:
```suggestion
{% if mysql_supports_innodb_large_prefix and mysql_cli_version.stdout is version('8.0', '<') %}
```

templates/my.cnf.j2
```diff
@@ -87,7 +87,7 @@ read_buffer_size = {{ mysql_read_buffer_size }}
 read_rnd_buffer_size = {{ mysql_read_rnd_buffer_size }}
 myisam_sort_buffer_size = {{ mysql_myisam_sort_buffer_size }}
 thread_cache_size = {{ mysql_thread_cache_size }}
-{% if '8.0.' not in mysql_cli_version.stdout %}
+{% if '8.' not in mysql_cli_version.stdout %}
```
Diegiwg, line 90:
```suggestion
{% if mysql_cli_version.stdout is version('8.0', '<') %}
```

tasks/secure-installation.yml
```diff
@@ -50,7 +62,7 @@
   with_items: "{{ mysql_root_hosts.stdout_lines|default([]) }}"
   when: >
     ((mysql_install_packages | bool) or mysql_root_password_update)
-    and ('5.7.' in mysql_cli_version.stdout or '8.0.' in mysql_cli_version.stdout)
+    and ('5.7.' in mysql_cli_version.stdout)
```
Diegiwg, line 65:
```suggestion
    and (mysql_cli_version.stdout is version('5.7', 'eq')
```

tasks/secure-installation.yml
```diff
@@ -41,6 +41,18 @@
...
+- name: Update MySQL root password for localhost root account (8.x).
+  ansible.builtin.shell: >
+    mysql -u root -NBe
+    "ALTER USER '{{ mysql_root_username }}'@'{{ item }}'
+     IDENTIFIED WITH caching_sha2_password BY '{{ mysql_root_password }}'; FLUSH PRIVILEGES;"
+  no_log: "{{ mysql_hide_passwords }}"
+  with_items: "{{ mysql_root_hosts.stdout_lines|default([]) }}"
+  when: >
+    ((mysql_install_packages | bool) or mysql_root_password_update)
+    and ('8.' in mysql_cli_version.stdout)
```
Diegiwg, line 54:
```suggestion
    and (mysql_cli_version.stdout is version('8.0', '>='))
```

tasks/secure-installation.yml
Diegiwg: @geerlingguy could you please give some attention to the alternative presented here?

I need to ensure support for MySQL 8.4 and at the moment, despite installing the service, the configuration is incorrect and the service does not start.

geerlingguy: It looks good — can you maybe file a PR based on this including your changes, and I can merge that once it's passing tests?

Diegiwg: @geerlingguy I think 8.4 support was implemented in PR https://github.com/geerlingguy/ansible-role-mysql/pull/561/files

Could you please validate?

PR #551, opened 2024-10-10
Allow mysql_config_include_files to contain raw content

frazew:
# Summary

In some cases, specifying a known src path in `mysql_config_include_files` is not possible / complicated and it may be much cleaner to provide raw file content to be sent. In that case, we need to use the `ansible.builtin.copy` module. We can extend the current implementation by simply filtering on whether `src` or `content` is set in each item in `mysql_config_include_files`

This PR also adds a label definition to the loop, that becomes necessary to avoid printing entire files with the default loop label

# Example

For example if you want to use `mysql_config_include_files` with a file stored inside another role (because that's where it's the most practical place to store it currently), you need to:
```yaml
mysql_config_include_files:
  - { src: "{{ role_path }}/../myrole/files/mysql/myconfig.cnf", force: yes }
```

Instead, for very simple configuration snippets it's cleaner to store them in the inventory / group_vars, and this PR then allows you to do that:
```yaml
mysql_config_include_files:
  - dst: myconfig.cnf
    content: |
      [mysqld]
      # ...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Whether the global my.cnf should be overwritten each time this role is run. Sett
 mysql_config_include_files: []
 ```
 
-A list of files that should override the default global my.cnf. Each item in the array requires a "src" parameter which is a path to a file. An optional "force" parameter can force the file to be updated each time ansible runs.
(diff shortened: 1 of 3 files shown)


Frazew:
I don't think this should be considered stale, it simply has not been reviewed yet but that's ok :smile:

frazew:
Still relevant! Let's keep it open if that's ok

Issue #544, opened 2024-06-24
support for rocky linux or el


oferchen:
hello,
it is possible to include support for rocky linux or rather enterprise linux family?

thanks

alainseys:
@oferchen  you can do this by changing the when statement

from
`- ansible.builtin.include_tasks: setup-RedHat.yml
  when: ansible_os_family == 'RedHat'`

to 
`- ansible.builtin.include_tasks: setup-RedHat.yml
  when: ansible_os_family == 'Rockey'`

but a better approuch should be like this

`- ansible.builtin.include_tasks: setup-RedHat.yml
  when: ansible_os_family in ['RedHat','Rockey'`]

PR #555, opened 2024-10-22
Add Rocky Linux configuration

SonOfWinter:
Refer to #544


SonOfWinter:
I was using an old version of ansible that didn't recognize RL.
I updated it and now everything works correctly.
My PR is now obsolete and I've closed it.

Issue #550, opened 2024-09-02
remove anonymous user in mysql_secure_installation fails with error 1045


schleicherdeha:
Here's the output of mysql_secure_installation:

[root@webserver~]# mysql_secure_installation

NOTE: RUNNING ALL PARTS OF THIS SCRIPT IS RECOMMENDED FOR ALL MariaDB
      SERVERS IN PRODUCTION USE!  PLEASE READ EACH STEP CAREFULLY!

In order to log into MariaDB to secure it, we'll need the current
password for the root user.  If you've just installed MariaDB, and
you haven't set the root password yet, the password will be blank,
so you should just press enter here.

Enter current password for root (enter for none): 
OK, successfully used password, moving on...

Setting the root password ensures that nobody can log into the MariaDB
root user without the proper authorisation.

You already have a root password set, so you can safely answer 'n'.

Change the root password? [Y/n] 
New password: 
Re-enter new password: 
Password updated successfully!
Reloading privilege tables..
 ... Success!

By default, a MariaDB installation has an anonymous user, allowing anyone
to log into MariaDB without having to have a user account created for
them.  This is intended only for testing, and to make the installation
go a bit smoother.  You should remove them before moving into a
production environment.

Remove anonymous users? [Y/n] 
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)
 ... Failed!
Cleaning up...

I did not set any variables in my playbook regarding mysql. Playbook had been running several times without errors. I didn't have this problem, when installing mysql/mariadb without ansible/awx.

Any help?

Issue #522, opened 2023-04-06
secure-installation fails on RHEL8 at remove test database.


bufanda:
The step to remove the test database during secure installations fails on a host that already is running for serveral month and the playbook was used to provision the host. subsequents runs now fail at this step with an Access denied.

Ansible Controller:
Red Hat Enterprise Linux 7.9
Ansible Core installation via package manager

Target Host
Red Hat Enterprise linux 8.7 

Playbook sets become to true.

The full Traceback
```
The full traceback is:
  File "/tmp/ansible_mysql_db_payload_034z8ww0/ansible_mysql_db_payload.zip/ansible/modules/database/mysql/mysql_db.py", line 383, in main
  File "/tmp/ansible_mysql_db_payload_034z8ww0/ansible_mysql_db_payload.zip/ansible/module_utils/mysql.py", line 81, in mysql_connect
    db_connection = mysql_driver.connect(**config)
  File "/usr/lib/python3.6/site-packages/pymysql/__init__.py", line 94, in Connect
    return Connection(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pymysql/connections.py", line 327, in __init__
    self.connect()
  File "/usr/lib/python3.6/site-packages/pymysql/connections.py", line 588, in connect
    self._request_authentication()
  File "/usr/lib/python3.6/site-packages/pymysql/connections.py", line 853, in _request_authentication
    auth_packet = self._read_packet()
  File "/usr/lib/python3.6/site-packages/pymysql/connections.py", line 676, in _read_packet
    packet.raise_for_error()
  File "/usr/lib/python3.6/site-packages/pymysql/protocol.py", line 223, in raise_for_error
    err.raise_mysql_exception(self._data)
  File "/usr/lib/python3.6/site-packages/pymysql/err.py", line 107, in raise_mysql_exception
    raise errorclass(errno, errval)
fatal: [db-003]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "ca_cert": null,
            "client_cert": null,
            "client_key": null,
            "collation": "",
            "config_file": "/root/.my.cnf",
            "connect_timeout": 30,
            "encoding": "",
            "ignore_tables": [],
            "login_host": "localhost",
            "login_password": null,
            "login_port": 3306,
            "login_unix_socket": null,
            "login_user": null,
            "name": [
                "test"
            ],
            "quick": true,
            "single_transaction": false,
            "state": "absent",
            "target": null
[...]

Issue #431, opened 2021-01-20
Access denied warnings with MariaDB 10.3 on Ubuntu 20.04


dawi:
After installing `MariaDB 10.3` on `Ubuntu 20.04` via this ansible role, 
I am getting the following warnings in `/var/log/mysql/error.log`:

```
2021-01-20 13:20:41 0 [Note] /usr/sbin/mysqld: ready for connections.
Version: '10.3.25-MariaDB-0ubuntu0.20.04.1-log'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306  Ubuntu 20.04
2021-01-20 13:20:41 8 [Warning] Access denied for user 'root'@'localhost' (using password: NO)
2021-01-20 13:20:41 9 [Warning] Access denied for user 'root'@'localhost' (using password: NO)
2021-01-20 13:20:41 10 [Warning] Access denied for user 'root'@'localhost' (using password: NO)
```

This warnings are caused during execution of `/etc/mysql/debian-start`:

```bash
…
MYSQL="/usr/bin/mysql --defaults-file=/etc/mysql/debian.cnf"
…
trap "" SIGHUP
(
  upgrade_system_tables_if_necessary;
  check_root_accounts;
  check_for_crashed_tables;
) >&2 &
…
```

The issue could be that passwords configured in `/etc/mysql/debian.cnf` and `/root/.my.cnf` don't match.

The errors do not occur if I set the `mysql_root_password` to `""` like this:

```
mysql_root_password:
mysql_packages:
  - mariadb-server-10.3
  - mariadb-client-10.3
```

But that's nothing I would want to use in production.

PierreRambaud:
I also have the same problem with debian 10.
I was wondering if adding the `-p {{ mysql_root_password  }}` is a good idea.
But, it should be easier and better to have a MySQL user which can run MySQL commands without using a password.

1977er:
Commenting out the empty `password =` lines in `debian.cnf` resolves any permission problems. (Ie. this role was unable to pass the `Remove MySQL test database` step due to allegedly missing permissions (ignoring correct /root/.my.cnf).)

MassiveHiggsField:
I hope i am not hijacking a different problem, but i had a similiar problem as described in the last comment here:

    TASK [geerlingguy.mysql : Remove MySQL test database.] *************************
[...]

PR #561, opened 2025-01-30
feat: use socket authentication as default for root

Al-thi:
This is an attempt to allow the root account to use socket authentication instead of password authentication.

Socket authentication is the new default for both MariaDB and MySQL on modern versions.

The current role configuration causes authentication problems for the root account, cf. issues:
* https://github.com/geerlingguy/ansible-role-mysql/issues/550
* https://github.com/geerlingguy/ansible-role-mysql/issues/522
* https://github.com/geerlingguy/ansible-role-mysql/issues/431
* https://github.com/geerlingguy/ansible-role-mysql/issues/421

I've had authentication problems too, with the following message on every `mysq_db` and `mysql_user` tasks:
```
2024-09-30 19:45:59,235 (item={'name': '[REDACTED]', 'password': '[REDACTED]', 'priv': '[MY_DB_NAME].*:ALL/mysql.*:SELECT,SHOW VIEW/*.*:BINLOG ADMIN'}) => {"ansible_loop_var": "item", "changed": false, "item": {"name": "[REDACTED]", "password": "[REDACTED]", "priv": "[MY_DB_NAME].*:ALL/mysql.*:SELECT,SHOW VIEW/*.*:BINLOG ADMIN"}, "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: (1698, \"Access denied for user 'root'@'localhost'\")"}
```

As this socket authentication is only effective on recent MySQL (≥ 8.4) / MariaDB (≥ 10.4) versions, this PR also refactors the version parsing by:
* extracting the interesting part of the `mysql --version` standard output
* using ansible `version` filter instead of substring detection (with `{{ 'something' in my_string }}` ansible variable)

Finally, as there was some linting warnings, I also included syntax fixes in the 3 last commits.

Please tell me what you think of this :)
Thank you for your time and your amazing work.
Regards.


brunick:
Hi @Al-thi , i've had this problem with my newest install as well, but i also got in some trouble trying you branch.

```
  - name: Run mysql --version
      ansible.builtin.command: 'mysql --version'
      register: mysql_cli_version
      changed_when: false
      check_mode: false
```

this does not apply to newest mariadb version, as those throw deprecation warning in their response of using mysql instead of mariadb:
```
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
mysql from 11.4.4-MariaDB, client 15.2 for Linux (x86_64) using  EditLine wrapper
```
could you update the script, to check for mysql vs mariadb and use mariadb if installed?
cheers

Al-thi:
@brunick Could you run the following command please:
```bash
ls -l $(which mysql)
```
I have seen many different configurations with some `mysql` binaries being symlinks to `mariadb`, other with plain `mysql` binaries which are in fact mariadb .... It's a mess really :sweat:

brunick:
> @brunick Could you run the following command please:
> 
> ```shell
> ls -l $(which mysql)
> ```
> 
> I have seen many different configurations with some `mysql` binaries being symlinks to `mariadb`, other with plain `mysql` binaries which are in fact mariadb .... It's a mess really 😓
> 
> EDIT: could you please try the latest version ? It should run `mariadb --version` instead of `mysql --version` for rockylinux.
Hey @Al-thi,
```
[root@mysql-server]# ls -l $(which mysql)
lrwxrwxrwx 1 root root 7 30. Okt 18:40 /usr/bin/mysql -> mariadb
```

brunick:
> @brunick did you try the latest version ?

not yet, i had to create a workaround to get my server up'n'running

i actually "only" had to add a root user with permissions to connect from 127.0.0.1 with a password.

molecule/default/converge.yml
```diff
@@ -11,6 +11,20 @@
       ansible.builtin.command: "mysql -u root -proot -e 'show databases;'"
       changed_when: false
 
-    - name: Make sure we can connect to MySQL via TCP.
+    - name: Make sure we can connect to MySQL via TCP on old MySQL version.
       ansible.builtin.command: "mysql -u root -proot -h 127.0.0.1 -e 'show databases;'"
```
geerlingguy, line 15: Can we still have a command to try connecting via socket separately to 'show databases' so we have this test for all versions of MySQL/MariaDB?

PR #571, opened 2025-04-10
fix: query caching features on mariadb

Al-thi:
Following on this comment in the code : https://github.com/geerlingguy/ansible-role-mysql/commit/e21a22afd48ce2f5e1707ed1875b33ecd162b396#commitcomment-154694645

This reintroduces `query_cache_*` variables for MariaDB users. This disappeared in #561 as I thought at the time it was OK to just remove these variables from MariaDB configuration. I was wrong :D

Thanks,
Regards


geerlingguy:
This is in 5.0.2. Thanks!

PR #594, opened 2026-03-11
Fix skipped removal of initial logs for Ubuntu

SuperFoo42:
Currently, the removal of the initial database logfile only works for Debian < 12, as the `distribution_version` comparison is tailored to Debian. 
For reasonably recent Ubuntu versions (e.g., 24), this does not work. However, Ubuntu's apt installation will also create initial log files that have to be removed. 

This PR addresses the issue by adding a check for Ubuntu. Furthermore, the apt installation will also create an initial ibdata-file that might interfere with custom user configuration of InnoDB like `innodb_page_size`. Therefore, this file is deleted as well.

### DIFF
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -4,7 +4,7 @@
   # become: true
 
   roles:
-    - role: geerlingguy.mysql
+    - role: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') }}"
 
   post_tasks:
     - name: Make sure we can connect to MySQL via Unix socket.
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -3,6 +3,10 @@
   hosts: all
   # become: true
 
+  vars:
+    mysql_config_include_files:
+      - src: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') }}/molecule/default/templates/override.cnf"
+
   roles:
     - role: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') }}"
 
--- a/molecule/default/templates/override.cnf
+++ b/molecule/default/templates/override.cnf
@@ -1,0 +1,2 @@
+[mysqld]
+innodb_page_size = 32K
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -2,10 +2,14 @@
 - name: Get MySQL version
   block:
     - name: Run mysql --version
-      ansible.builtin.command: '{{ mysql_daemon }} --version'
+      ansible.builtin.command: "{{ mysql_daemon }} --version"
       register: mysql_cli_version
       changed_when: false
       check_mode: false
+
+    - name: Set MariaDB flag.
+      ansible.builtin.set_fact:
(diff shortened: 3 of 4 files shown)


geerlingguy:
Looks fine, but please try to fix the lint issues before merge.

SuperFoo42:
Linter should be satisfied now.
I noticed the molecule tests are failing. Same for me locally, but it seems to be unrelated to the change.

geerlingguy:
The last CI run was passing as of March 10:

https://github.com/geerlingguy/ansible-role-mysql/actions/runs/22932344253

The error in CI seems to be:

```
TASK [geerlingguy.mysql : Ensure MySQL is started and enabled on boot.] ********
  [ERROR]: Task failed: Module failed: Unable to start service mysql: Job for mysql.service failed because the control process exited with error code.
  See "systemctl status mysql.service" and "journalctl -xeu mysql.service" for details.
  
  Origin: /home/runner/work/ansible-role-mysql/ansible-role-mysql/geerlingguy.mysql/tasks/configure.yml:91:3
  
  89   tags: ['skip_ansible_galaxy']
  90
  91 - name: Ensure MySQL is started and enabled on boot.
       ^ column 3
  
  fatal: [instance]: FAILED! => {"changed": false, "msg": "Unable to start service mysql: Job for mysql.service failed because the control process exited with error code.\nSee \"systemctl status mysql.service\" and \"journalctl -xeu mysql.service\" for details.\n"}
```

SuperFoo42:
Sorry it took a bit longer to find time again.

It turned out it is a bit more complicated, because deleting ibdata leaves tablespace metadata still in place, leading to MySQL refusing to start. The main problem, however, was that I could not reproduce this problem first. The molecule test runs against the galaxy role, not against the local repository. @geerlingguy Is that intended? I changed that in [b3f8158](https://github.com/geerlingguy/ansible-role-mysql/pull/594/commits/b3f815836e88f8bd1cf087fb8ab8e5965cd2e2b0).

The new behavior is to delete the data directory entirely, recreate it, and later reinitialize. I changed the recreation permissions in accordance with the recommendation from [MySQL](https://dev.mysql.com/doc/mysql-secure-deployment-guide/5.7/en/secure-deployment-permissions.html).
I also removed the OS checks between Debian and Ubuntu entirely, as I could reproduce the problem on Debian 13 as well. This likely affects the entire deb install path.

93jun:
Hi, I may have run into a similar Ubuntu 24.04 install-path issue while testing this role locally.

The run reached:

```text
TASK [mysql : Ensure MySQL packages are installed.]
```

and then `apt`/`dpkg` failed while configuring `mysql-server-8.0`:

```text
/usr/bin/apt-get ... install mysql-server=8.0.46-0ubuntu0.24.04.3 failed
E: Sub-process /usr/bin/dpkg returned an error code (1)
dpkg: error processing package mysql-server-8.0 (--configure):
 installed mysql-server-8.0 package post-installation script subprocess returned error exit status 1
```

The same log also showed MySQL starting during package configuration and then failing to shut down cleanly:

```text
mysqld will log errors to /var/log/mysql/error.log
mysqld is running as pid ...
Error: Unable to shut down server with process id ...
```

This looks related to the initial log/data-file handling discussed in this PR. I do not have a smaller standalone reproduction yet, but I wanted to share the log excerpt in case it is useful.

PR #595, opened 2026-03-23
fix: support custom user creation on MySQL 8.4+

Diegiwg:
> [!NOTE]
> The work on this PR was sponsored by [Devopness](https://github.com/devopness/devopness) 

## Problem

Creating custom users via the `mysql_users` variable fails on MySQL 8.4+ with authentication errors.

## Root Cause

MySQL 8.4+ disables `mysql_native_password` by default. The role's user creation logic still uses the traditional password-based approach, which no longer works.

## Solution

- Use `caching_sha2_password` plugin for MySQL 8.0+
- Keep traditional password auth for MySQL < 8.0 and MariaDB
- Fully backward compatible

## Testing

- MySQL 8.4 on Ubuntu 24.04
- MySQL 8.0 on Ubuntu 24.04

## References

- [MySQL 8.4 Native Pluggable Authentication (Deprecation)](https://dev.mysql.com/doc/refman/8.4/en/native-pluggable-authentication.html)
- [Caching SHA-2 Pluggable Authentication](https://dev.mysql.com/doc/refman/8.4/en/caching-sha2-pluggable-authentication.html)

### DIFF
--- a/tasks/users.yml
+++ b/tasks/users.yml
@@ -1,6 +1,29 @@
 ---
-- name: Ensure MySQL users are present.
-  mysql_user:
+# For MySQL >= 8.0, use caching_sha2_password since mysql_native_password
+# is disabled by default in MySQL 8.4+ and will be removed in MySQL 9.0.
+# See: https://dev.mysql.com/doc/refman/8.4/en/native-pluggable-authentication.html
+- name: Ensure MySQL users are present (MySQL >= 8.0).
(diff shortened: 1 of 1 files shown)


Diegiwg:
@geerlingguy could you review this when you have a chance?

This fixes custom user creation (`mysql_users` variable) on MySQL 8.4+.

Thank you for the amazing work on this role (and other role's)

PR #607, opened 2026-08-29
Separate the client binary and engine flavour from the service name

mcowser-p:
On Enterprise Linux running **real MySQL** (not MariaDB), this role cannot complete: the systemd unit is `mysqld` while the client binary is `mysql`, and `mysql_daemon` has to be both at once.

## The three meanings

`mysql_daemon` is currently used as:

1. the **systemd unit** — `service: name={{ mysql_daemon }}`
2. the **client binary** — `secure-installation.yml:25`, `command: {{ mysql_daemon }} -NBe "..."` (six call sites)
3. the **engine flavour** — `mysql_daemon == 'mariadb'` / `== 'mysql'` (seven tests)

That is coherent wherever the three share a name: MariaDB everywhere, and MySQL on Debian, where the unit really is `mysql`. On EL + MySQL there is no value that works:

| value | client | flavour | unit |
|---|---|---|---|
| `mariadb` (EL default) | not installed | wrong engine | wrong |
| `mysqld` | runs the **server** with client flags | matches neither test | correct |
| `mysql` | correct | correct | wrong |

The `mysqld` case is the unpleasant one: `mysqld -NBe "DELETE FROM mysql.user ..."` tries to *start a database* with client arguments, blocks for ~100 seconds, then exits 1.

## The change

Adds `mysql_client` and `mysql_flavour`, both defaulting to `mysql_daemon`, so **every existing playbook is unaffected**. Client invocations use `mysql_client`; flavour comparisons use `mysql_flavour`.

Callers who need the split set it explicitly:

```yaml
mysql_daemon:  mysqld
mysql_client:  mysql
mysql_flavour: mysql
```

## Testing

Found and verified running MySQL 8.0.46 on AlmaLinux 9 and MySQL 8.4 on Ubuntu 26.04. With the change, a host that had failed at `mariadb --version` on every run now completes the role cleanly. Debian is unaffected either way, which is why this has stayed hidden.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mcowser-p:
Closing — I opened this prematurely. Keeping the change on a private fork for now; will resubmit if it's useful upstream later. Apologies for the noise.

